PR #6, opened 2015-05-12
Use `date_to_rfc822` filter for date

pathawks:
Jekyll has a [built-in filter](https://github.com/jekyll/jekyll/blob/b41ddbb8a92934e9b02ba0b4dd715496a519e59e/lib/jekyll/filters.rb#L83-L95) just for this sort of thing. :grinning:


benbalter:
Nice. Just [confirmed that RFC822 is the proper spec](https://validator.w3.org/feed/docs/error/InvalidRFC2822Date.html), which I'm dropping here for linkabliity.

@pathawks if you can resolve the merge conflict, I'd be glad to merge. Thanks again! :turtle:

pathawks:
> if you can resolve the merge conflict, I'd be glad to merge.

:heavy_check_mark:

PR #7, opened 2015-05-12
`xml_escape` ALL THE THINGS!

pathawks:


benbalter:
Nice catch. Would you be able to add a quick test (Looking just at post content is probably sufficient for our purposes) to make sure we don't break this going forward?

pathawks:
> Would you be able to add a quick test

Yes; give me just a couple days. I will have more time after this week is over.

pathawks:
Closing because #24

PR #8, opened 2015-05-12
Remove unused namespaces

pathawks:
Since Jekyll does not do anything with comments, let alone publish dynamically updated comment feeds, I can't imaging we'll *ever* need `wfw` or `slash`

The others I removed because they are not currently being used. It is easy enough to put them back if and when the need ever arises.


benbalter:
Looks good to my uneducated brain, but XML namespaces aren't my area of expertise, so I'm going to leave this open for a bit to give some other, much smarter folks the chance to weigh in if they'd like.

pathawks:
For reference, the [Jekyll site template](https://github.com/jekyll/jekyll/blob/master/lib/site_template/feed.xml#L5) uses simply:

```xml
<rss version="2.0" xmlns:atom="http://www.w3.org/2005/Atom">
```

XhmikosR:
Those are just unused.

benbalter:
Thanks @pathawks @XhmikosR :tada:

PR #9, opened 2015-05-12
Remove references to English

pathawks:
Not all sites will use English exclusively. Since we do not know that this information is English, we should not make that claim.

It does not seem this impacts validation, as the [Jekyllrb.com sitemap](https://github.com/jekyll/jekyll/blob/master/lib/site_template/feed.xml) includes no such information.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -12,13 +12,12 @@ permalink: /feed/index.xml
 	xmlns:slash="http://purl.org/rss/1.0/modules/slash/"
 	>
 <channel>
-    <title xml:lang="en">{{ site.name }}</title>
+    <title>{{ site.name }}</title>
     <atom:link type="application/atom+xml" href="{{ url_base }}/feed/" rel="self"/>
     <link>{{ url_base }}</link>
     <pubDate>{{ site.time | date: "%a, %d %b %Y %H:%M:%S %z" }}</pubDate>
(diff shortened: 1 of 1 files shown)


benbalter:
> It does not seem this impacts validation

[Looks like that's true](https://validator.w3.org/feed/check.cgi?url=jekyllrb.com%2Ffeed.xml). Is there any value in having it at all?

Also @pathawks are you able to resolve the merge conflict?

pathawks:
> Also @pathawks are you able to resolve the merge conflict?

Resolved :smile: 

> Is there any value in having it at all?

Oh sure, if we actually knew what language the content was. Some programs will offer to translate content that is not in the users default language.

pathawks:
> Also @pathawks are you able to resolve the merge conflict?

Resolved. :smile: 

> Is there any value in having it at all?

Oh sure, if we have a way of actually knowing the language of the content.

Some programs may offer to automatically translate content that is not in the users preferred language.

parkr:
:+1:

XhmikosR:
How about adding an option for this? While not everything is in English, specifying the language code should offer maximum flexibility.

pathawks:
I can see maybe a `site.language` and/or a `post.language` setting, but for now I think it's probably best to keep simple things simple on the road to first stable version.

XhmikosR:
I hear you, but I'd like to have an issue to track this since I'd like to be as close to the specs as possible.

I'll make a new issue for this.

pathawks:
> I'll make a new issue for this.

:+1:

Issue #2, opened 2015-05-12
What feed format is best?


benbalter:
RSS 1.0? 2.0? Atom? Is there any difference in use or reader adoption?

daviddarnes:
I've found that RSS seems to be the best if you're wanting to use this with iTunes, if that helps?

daviddarnes:
RSS 2*

pathawks:
> I've found that RSS 2 seems to be the best if you're wanting to use this with iTunes, if that helps?

I have found this as well, when publishing a podcast.

For all other use cases, I prefer authoring Atom 1.0. It feels like, because it is newer, it had the benefit of hindsight and was able to smooth over some of the rougher edges of RSS.

envygeeks:
The problem is that the biggest data provider on the planet uses Atom to feed it's system of news (Google) but that said they are also smart enough to reformat RSS as Atom before feeding it (I wish more companies normalized like that.)

parkr:
There is a gigundous amount of prior art here, both as gems and in blog post tutorial form. The top 10 Google results:

- [batsov](http://batsov.com/articles/2011/04/24/add-atom-feed-to-jekyll/) recommends Atom.
- [snaptortoise/jekyll-rss-feeds](https://github.com/snaptortoise/jekyll-rss-feeds/blob/master/feed.xml) seems to use RSS 2 / Atom.
- [agelber/jekyll-rss](https://github.com/agelber/jekyll-rss/blob/master/rssgenerator.rb#L40) uses RSS 2.
- [Drew Inglis](http://blog.drewinglis.com/2013/03/18/adding-rss.html) recommends RSS 2 / Atom.
- [Our good friend Joel Glovier](http://joelglovier.com/writing/rss-for-jekyll/) also recommends RSS 2 / Atom.
- [Mikel with the cool last name](http://kmikael.com/2013/05/23/adding-an-rss-feed-to-a-jekyll-site/) recommends RSS 2 / Atom.
- [Mr. Landau here](http://natelandau.com/jekyll-rss-feed-template/) recommends RSS 1, but it's a year old.
- [Dave Coyle](https://davecoyle.com/tech-notes/jekyll-templates-for-atom-rss/) offers templates for both Atom 1.0 and RSS 2.0.
- [Our dear friend and long-time Jekyll user Zanshin](http://zanshin.net/2013/11/08/how-to-build-an-rss-feed-for-jekyll/) recommends Atom.

[...]

PR #10, opened 2015-05-12
Fix content type

pathawks:
The feed is RSS2, it should not claim to be Atom.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -10,7 +10,7 @@
 	>
 <channel>
     <title>{{ site.name }}</title>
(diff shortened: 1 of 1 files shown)


benbalter:
@pathawks I believe you are right, but going to let this sit for a bit, pending the discussion over in https://github.com/jekyll/jekyll-rss-feed/issues/2#issuecomment-101388927, so as not to mislead anyone before that decision lands.

XhmikosR:
This is definitely right for the current code.

benbalter:
Merging this for now, but see #2.

PR #12, opened 2015-05-12
Remove front matter from template

pathawks:
Since we're never actually reading the front matter from the template, if we want to specify a permalink, we need to do it in Ruby.

Previously, the XML output file just contained the YAML front matter :-1:


parkr:
:+1:

benbalter:
Ha. Nice catch.

PR #13, opened 2015-05-12
Do not strip newlines from post content

pathawks:
Since posts may contain code blocks, or some other content with pre-formatted whitespace, we *do not* want to strip whitespace from post content.

This pull request strips whitespace from the template *before* it is rendered, rather than after. This will keep all the whitespace from the post content intact.


parkr:
Example? We should add a test for this.

pathawks:
Test has been added

benbalter:
Nice. Thanks for adding the test @pathawks (and for all the contributions!)

pathawks:
> Thanks for adding the test @pathawks (and for all the contributions!)

Thank you @benbalter; I've been waiting for a plugin like this for a while! :beers:

Issue #4, opened 2015-05-12
What should the path to the RSS feed be?


benbalter:
Theoretically, as long as it's exposed in the page metadata, it doesn't matter to computers.

I'm personally a fan of `/feed/` (rendered as `feed/index.xml`), because pretty permalinks. 

I've also seen `rss.xml`, `atom.xml`, and `feed.xml`, which seem less intuitive to me (going to a site, I can guess `/feed`, just as I might guess `/about`, but `/atom.xml` seems a bit more opaque).

The other issue to consider, most servers default configuration will look to `index.xml` after `index.html` (and serve the proper doctype), but there are some edge cases where it may not (GitHub Pages supports XML indexes for pretty permalinks).

Last, it feels silly, but we could theoretically say "all of the above" and render the same template into multiple spots (redirect-from isn't an option here, because redirect headers).

pathawks:
I don’t know how important a pretty permalink is for a feed. How often would a user need to type the URL? Once?

Would respect to avoiding config bloat, it seems reasonable to allow the user to specify a path for the feed. If migrating to Jekyll from somewhere else, it would be A Good Thing to avoid breaking the old feed.

parkr:
I see `/feed.xml` on most sites. As long as the header is specified on the main page of the site, most RSS readers will figure out the location. I'd be :+1: for `/feed.xml` due to the incidence of its use on blogs I visit. What does WordPress do?

parkr:
> If migrating to Jekyll from somewhere else, it would be A Good Thing to avoid breaking the old feed.

This should be mitigated by the headers in the HTML, yeah?

pathawks:
> This should be mitigated by the headers in the HTML, yeah?

The way that sort of autodiscovery works is, when I subscribe to a feed in my feed reader, the feed reader will fetch the URL that I entered and (1) if it is a feed, it will store that URL as the canonical location of the feed or (2) if it is an HTML page, it will find the URL of the feed from the autodiscovery tags, and then store that URL as the canonical location of the feed.

Either way, what the feed reader remembers is just the URL of the feed. If that URL 404s, none of the feed readers that I have used have been sophisticated enough to go looking to see if there is a new feed in a new location. They will not check the autodiscovery headers of the HTML. They will (at best) report the 404 or (at worst) just silently forget about the feed.

> What does WordPress do?

`/?feed=rss2` or `/feed/` or `/feed` - RSS 2.0

[...]

PR #17, opened 2015-05-12
Remove line that did nothing

pathawks:
I'm sorry, I made a change that didn't actually do what it looked like it did.

In line 46, the new page is called `feed.xml`; setting a permalink after that doesn't actually change anything.

### DIFF
--- a/lib/jekyll-rss-feed.rb
+++ b/lib/jekyll-rss-feed.rb
@@ -46,7 +46,6 @@ module Jekyll
       site_map = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", "feed.xml")
       site_map.content = File.read(source_path)
       site_map.data["layout"] = nil
-      site_map.data["permalink"] = "/feed/index.xml"
       site_map.render(Hash.new, @site.site_payload)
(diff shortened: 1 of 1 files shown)


parkr:
?

pathawks:
If you look at the `spec/dest/`, `feed/index.xml` does *not* exist, but `feed.xml` does.

Setting the `permalink` does not actually affect where the file is created.

pathawks:
I have updated the pull request so that it _actually_ renders the feed at `feed/index.html` as expected.

XhmikosR:
I've also updated README.md in #18.

XhmikosR:
One thing you missed @pathawks https://github.com/jekyll/jekyll-rss-feed/blob/master/lib/feed.xml#L9

pathawks:
With @benbalter's [comment on #4](https://github.com/jekyll/jekyll-rss-feed/issues/4#issuecomment-101658550) in mind, I have reverted the change that rendered the feed at `/feed/index.xml`.

It seems like `/feed.xml` is the place for the feed.

benbalter:
:metal: thanks again @pathawks :jack_o_lantern:

PR #18, opened 2015-05-13
Update README.md

XhmikosR:


README.md
```diff
@@ -21,11 +21,11 @@ gems:
 
 ## Usage
 
-The plugin will automatically generate a RSS feed at `/feed/`. No configuration necessary.
+The plugin will automatically generate an RSS feed at `/feed.xml`. No configuration necessary.
 
 ## Contributing
 
-1. Fork it ( https://github.com/jekyll/jekyll-rss-feed/fork )
```
XhmikosR: BTW, the reason I removed these spaces is that even though (I assume) they were added to match the space the other brackets have, that space is `margin` set via CSS.

XhmikosR:
Added `rss_limit` too.

Do not merge this yet, I want to squash the patches first.

Also, this reminds me, how about having the config option under `rss` like

```yml
rss:
  limit: 10
```

benbalter:
@XhmikosR is there a corresponding pull request to add the RSS limit logic? At first glance I'd be :-1: to add that config setting, let alone three, as outlined over in https://github.com/jekyll/jekyll-rss-feed/issues/5#issuecomment-101393563.

XhmikosR:
@benbalter: what Pull Request?

https://github.com/jekyll/jekyll-rss-feed/blob/master/lib/feed.xml#L15

benbalter:
Ahh. Didn't realize that slipped in there.

XhmikosR:
I tweaked the changes, please have a look now.

benbalter:
Nice. Thanks @XhmikosR!

XhmikosR:
No problem, I was expecting for such a plugin for years since I wanted a common way to generate the feeds apart from using my solution.

PR #21, opened 2015-05-13
Validate the resulting feed

### LINKED ISSUE #16: Validate feed
I wonder if we could use something like [alexdunae/w3c_validators](https://github.com/alexdunae/w3c_validators) to let Travis validate our feed.

This would take a lot of the guess work out of a particular change being ok or not.

benbalter:
Both by parsing the RSS for expectations, and by calling the W3 validation API.

To note, there are some gems out there, but they haven't been updated in 3+ years, and given one call, things were simple enough to implement with a post request + nokigiri.

Fixes #16


parkr:
Nifty!

pathawks:
:cool:

Issue #23, opened 2015-05-13
Use existing Post GUID if present


pathawks:
I believe that some importers [will save a posts old GUID in the YAML front matter](https://github.com/jekyll/jekyll-import/blob/master/lib/jekyll-import/importers/wordpress.rb#L128).

It would be a Good Thing for us to make use of this data, if present. If a Jekyll feed does not use the same GUIDs as the feed generated by the old CMS, feed readers will mistakenly think that a feed is full of new items and display all items as unread. Using the same IDs prevents this.

pathawks:
And [we're already doing this](https://github.com/jekyll/jekyll-rss-feed/blob/master/lib/feed.xml#L22).

Fantastic.

PR #24, opened 2015-05-14
Use Atom template for feed

### LINKED ISSUE #2: What feed format is best?

benbalter:
This pull request updates `/feed.xml` to serve Atom, rather than RSS, per the discussion over in #2. 

The pull request also updates the documentation, continuing to use the term RSS at times, because as @pathawks noted, although not technically accurate, it is often used a generic term to describe blog feeds. :sparkles: SEO :sparkles: 

The Atom feed validates against W3 with no warnings.

I also wanted to call out a few optional, common site-level and post-level configuration variables, which are baked into the template:

### Optional configuration options

The plugin will automatically use any of the following configuration variables, if they are present in your site's `_config.yml` file.

* `name` - The title of the site, e.g., "My awesome site"
* `url` - The URL to your site, e.g., `http://example.com`. If none is provided, the plugin will try to use `site.github.url`.
* `author` - Your name, e.g., "Dr. Jekyll"

### Optional front matter

The plugin will use the following post metadata, automatically generated by Jekyll, which you can override via a post's YAML front matter:

* `date`
* `title`
* `excerpt`
* `id`
* `category`
* `tags`

Additionally, the plugin will use the following values, if present in a post's YAML front matter:

* `author` - The author of the post, e.g., "Dr. Jekyll". If none is given, feed readers will look to the feed author as defined in `_config.yml`

Fixes #2.


XhmikosR:
So, do you plan to rename the plugin to jekyll-atom-feed too?

I still believe offering the 2 is the best solution for Jekyll users.

benbalter:
> So, do you plan to rename the plugin to jekyll-atom-feed too?

Opened https://github.com/jekyll/jekyll-rss-feed/issues/27 to discuss.

lib/feed.xml
```diff
@@ -1,26 +1,57 @@
...
-<channel>
+{% if site.url %}
+  {% assign url_base = site.url | append: site.baseurl %}
+{% else %}
+  {% assign url_base = site.github.url %}
+{% endif %}
+<feed xmlns="http://www.w3.org/2005/Atom">
+  <generator uri="http://jekyllrb.com" version="{{ jekyll.version}}">Jekyll</generator>
+  <link href="{{ url_base }}{{ page.url }}" rel="self" type="application/atom+xml" />
+  <link href="{{ url_base }}/" rel="alternate" type="text/html"/>
```
pathawks: Is the slash after `{{ url_base }}` necessary?

lib/feed.xml
```diff
@@ -1,26 +1,57 @@
...
+  {% endif %}
+
+  {% for post in site.posts %}
+    <entry>
+      <title>{{ post.title | xml_escape }}</title>
+      <link href="{{ url_base }}{{ post.url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
+      <published>{{ post.date | date_to_xmlschema }}</published>
+      <updated>{{ post.date | date_to_xmlschema }}</updated>
+
+      <id>{{ url_base }}{{ post.id }}</id>
```
pathawks: Could we allow a site to override this, maybe with `guid` if present?

This will allow sites to move to Jekyll without breaking their existing feeds for users.

See #23

lib/feed.xml
```diff
@@ -1,26 +1,57 @@
...
-	>
-<channel>
+{% if site.url %}
+  {% assign url_base = site.url | append: site.baseurl %}
+{% else %}
+  {% assign url_base = site.github.url %}
+{% endif %}
+<feed xmlns="http://www.w3.org/2005/Atom">
+  <generator uri="http://jekyllrb.com" version="{{ jekyll.version}}">Jekyll</generator>
+  <link href="{{ url_base }}{{ page.url }}" rel="self" type="application/atom+xml" />
```
pathawks: Can we remove the space before the `/>`? Line 10 does not include such a space.

Micro-optimization?

lib/feed.xml
```diff
@@ -1,26 +1,57 @@
...
+      {% if post.category %}
+        <category term="{{ post.category }}" />
+      {% endif %}
+
+      {% for tag in post.tags %}
+        <category term="{{ tag }}" />
+      {% endfor %}
+
+      {% if post.excerpt %}
+        <summary>{{ post.excerpt | strip_html | strip_newlines }}</summary>
```
pathawks: `post.excerpt` should be escaped.

Per [Atom spec](http://atomenabled.org/developers/syndication/#text), all text should be escaped.

lib/feed.xml
benbalter: Wouldn't `post.id` be that GUID? (or am I not understanding?)

lib/feed.xml
benbalter: Added a space, as all the other tag closures had a space, but not strongly opinionated.

lib/feed.xml
```diff
@@ -1,26 +1,57 @@
...
-</rss>
+  {% endif %}
+
+  {% if site.description %}
+    <subtitle>{{ site.description }}</subtitle>
+  {% endif %}
+
+  {% if site.author %}
+    <author>
+      <name>{{ site.author }}</name>
```
sondr3: Should this just be `{% site.author %}` or should it be `{% site.author.name %}`? Because I use my `_config.yml` files with a `author` block that contains name, email, twitter and so on. I think most people have their own different ways of doing this and I think it'd be great if we had one more officially supported way of doing this to help with other plugins as well.

benbalter: @sondr3 Great call. I'd love to get there eventually (with the simple fallback), but it was a bit more than I wanted to tackle for a first pass. Opened https://github.com/jekyll/jekyll-rss-feed/issues/28 to discuss.

benbalter:
Thanks all for the feedback. Sounds like we've got a few follow up discussion to tackle in new issues, but things are at least solid enough here to merge, for now, and we can continue to improve the template.

PR #25, opened 2015-05-14
Add feed_meta tag

### LINKED ISSUE #20: Liquid tag to add meta tags to head
Thinking something like `{{ feed_meta }}` which would output:

```html
<link rel="alternate" type="application/rss+xml" title="RSS" href="/feed.xml" />
```

benbalter:
Such that `{% feed_meta %}` will output `<link type="application/atom+xml" rel="alternate" href="http://example.org/feed.xml" />`

Fixes #20

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,6 +27,10 @@ Optional configuration options:
 
 * `rss_limit`: number of posts to be included in the feed; default `nil`
 
+### Meta tags
+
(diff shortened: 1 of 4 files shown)


XhmikosR:
I'd say you can squash those patches.

Apart from that, the end tag isn't really needed in HTML 5. Does anybody still care about HTML 4?

pathawks:
HTML 5 as XHTML is a thing.

Yes, keep the self-closing tag.

XhmikosR:
Live in 1990 much?

PR #30, opened 2015-05-18
add trailing slash back in

benbalter:
Revert f61d75749a45aebfa53293cee046f2896cde9d5d which removed the trailing slash in the `id`.

Without it, we get the validation warning `Validation warning: Identifier "http://example.org" is not in canonical form (the canonical form would be "http://example.org/") on line 7 column 22`.

/cc https://github.com/jekyll/jekyll-rss-feed/pull/24


pathawks:
:+1: You beat me to it :stuck_out_tongue:

PR #31, opened 2015-05-18
Better author support

### LINKED ISSUE #28: Better author support
Allow defining a site/post author as a simple key value pair (`name: Ben`) or, as an object, e.g.:

```yml
author:
  name: Ben
  uri: "http://ben.balter.com"
  email: ben@example.com
```

In the first case, we'd just set `author.name` to `Ben`, and in both cases, only output the field (which is supported within the atom spec) if present.

pathawks:
Fixes #28


benbalter:
Ahh nice. Checked the docs, and `name` is required, so there is either `site.author.name`, or it's not a valid object. 

One thought, how do we want to handle if no author is specified, as either a string or an object? Right now, we output an empty node.

Last, would you mind updating the docs with this behavior? Awesome stuff.

pathawks:
> Ahh nice. Checked the docs, and `name` is required, so there is either `site.author.name`, or it's not a valid object.
> 
> One thought, how do we want to handle if no author is specified, as either a string or an object? Right now, we output an empty node.

So if *no* author information is provided, we don't output an `<author>` tag, and that's ok.

If *some* author information is provided without a name (like URL or email), creating an `<author>` tag without a name is invalid.

My first thought is that this is a problem best solved with documentation. We could just make it clear that, if your site/post has an author, you must provide the author's name.

:grey_question:

benbalter:
> So if no author information is provided, we don't output an <author> tag, and that's ok.

Missed the containing `{% if site.author %}`. :+1:

pathawks:
I'm bad at documentation (which is to say, I suppose, that I am a lazy programmer).

Can somebody take a look and make some suggestions?

pathawks:
For the feed author, if nothing is provided, should we try to fallback to `site.github.owner`, similar to what is done for `site.url`?

PR #32, opened 2015-05-18
Escape every string before output

pathawks:
Whether malicious or accidental, not escaping strings before outputting them to the feed opens the door for a user to totally bork their feed.

This pull request tries to escape everything that might be a string before outputting to the feed.


benbalter:
Nice. And a test! Thanks again for the pull request.

PR #33, opened 2015-05-18
Refactor feed.xml to use prepend

pathawks:
I know this is the worst kind of PR: the only reason I did this is because it the convention used for [**jekyll-sitemap**](https://github.com/jekyll/jekyll-sitemap/commit/88aad5c8ca55644ff2afe8c74847c3d584a0289a), but @parkr's a smart guy, I'm sure there is a reason behind the convention...?


benbalter:
>  @parkr's a smart guy

Can't argue with that. :+1: Thanks again.

PR #35, opened 2015-05-19
Geotagging

pathawks:
This pull request adds basic geotagging support for posts.

It uses [GeoRSS](http://www.georss.org/simple.html) to add the information to the feed, which is supported by feed readers like [Flipboard](https://about.flipboard.com/rss-spec/).


parkr:
This feels out of scope for this project. That said, I didn't write it so I don't know if it's really in the scope. Maybe @benbalter has an idea.

pathawks:
I understand where you're coming from. However, this is a feature that both publishers and consumers use; it would be a shame if a site could not use this plugin at all just because they needed this one feature.

benbalter:
Cool feature, and nicely implemented. Can you share a bit more about your use case (or how others might use it)? Just trying to get a sense for how it's used, both as a reader and as a publisher.

pathawks:
For a publisher: If a blog post is about a specific location (ie: travel blog, restaurant review...), that location can be tied to the post. I have used this in the past to show a map of recent posts, just by handing the Atom feed to the Google Maps API.

As far as the reader goes, I don't know. Maybe there's no benefit to them.

Maybe this is one of those things that was the New Hotness in 2008 and I haven't reevaluated since then.

FWIW, [WordPress.com supports this sort of geotagging](https://en.support.wordpress.com/geotagging/), but with little explanation of why one would do this.

pathawks:
I'm closing this for now.

As much as I love the idea of this feature, I admit that there isn't much obvious benefit to readers.

I would love to revisit this in the future if somebody finds a compelling use case. The nice part about doing a feed as a Gem is that it is easy to add features in the future.

Thank you @parkr and @benbalter for asking the tough questions.

Issue #22, opened 2015-05-13
Replace URLs with full URLs in post.content


XhmikosR:
http://validator.w3.org/feed/docs/warning/ContainsRelRef.html

Personally, I hacked around this https://github.com/mpc-hc/mpc-hc.org/blob/master/source/_plugins/fix_rss.rb in order to remove unneeded stuff from the feed.

PR #36, opened 2015-05-19
Add a `feedsafe` filter

pathawks:
This is a first step toward addressing #22

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -33,7 +33,7 @@
       <updated>{{ post.date | date_to_xmlschema }}</updated>
 
       <id>{{ post.id | prepend: url_base | xml_escape }}</id>
-      <content type="html">{{ post.content | markdownify | xml_escape }}</content>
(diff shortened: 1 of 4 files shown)


benbalter:
What's the downside of having classes in an Atom feed? Same with protocol-relative URLs?

pathawks:
I'm not sure classes would have much meaning without whatever stylesheet goes with the site, but I suppose they're probably not hurting much either.

I don't know how offline feed readers deal with protocol-relative URLs; perhaps filtering them is unnecessary? I guess we could ignore them until it causes a real world problem.

pathawks:
Also of note, WordPress - in addition to filtering content of the feed - [includes `xml:base`](https://github.com/WordPress/WordPress/blob/master/wp-includes/feed-atom.php#L72) so that relative permalinks can still be relative to something.

pathawks:
Closing this, as #43 seems to be good enough.

PR #37, opened 2015-05-20
Render post content prior to feed generation

benbalter:
It appears, right now, `post.content` is still markdown at the time the feed is processed. Ideally, we'd want this to be HTML. 

@parkr is it even possible to get the Markdown content as a generator? (or would we need to reimplement the renderer to render it ourselves)?

Fixes https://github.com/jekyll/jekyll-feed/issues/38.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -27,13 +27,13 @@
 
   {% for post in site.posts %}
     <entry>
-      <title>{{ post.title | xml_escape }}</title>
+      <title>{{ post.title | markdownify | strip_html | strip_newlines | xml_escape }}</title>
       <link href="{{ url_base }}{{ post.url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
       <published>{{ post.date | date_to_xmlschema }}</published>
       <updated>{{ post.date | date_to_xmlschema }}</updated>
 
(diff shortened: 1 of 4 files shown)


pathawks:
This seems somewhat related: jekyll/jekyll#2901

pathawks:
May also want to add a test to check that Liquid is rendered in post content (currently it is not).

benbalter:
Moved things over to `markdownify`, rather than vanilla `page.content` to at least get things working so that we can test this in the while, but 100% agree, ideally we'd somehow get the rendered post content, including liquid, prior to the layout. 

@parkr if you have any creative solutions that don't require changes to core, I'm all :ear:s! :smile:

swanson:
I'm having trouble moving my site over to the feed plugin. I think this is the same issue. I have some posts with, e.g.

```
![]({{site.url}}/static/some-image.jpg)  
```

I want the full URL to show up in the feed. The output in the generated feed is:

```
<img src="%7B%7Bsite.url%7D%7D/static/some-image.jpg" alt="">
```

Any ideas?

pathawks:
@swanson @nicerobot This pull request has been merged and is more related to rendering Markdown content, where your questions are more about rendering Liquid template tags.

If you are still having these issues with the latest version of this plugin, please open a new issue.

PR #39, opened 2015-05-20
XML escape the ID field

benbalter:
Literally the only field that wasn't XML escaped (and the only one preventing my test feed from validating).


pathawks:
I suppose that's my mistake. :sheep:

PR #41, opened 2015-05-20
Limit feed to 10 posts

### LINKED ISSUE #40: Limit RSS feed entries
Sicking Jekyll Feed loose on my personal site, with 100+ posts, resulted in a 1.1MB file, which is a bit silly for a feed... it shouldn't be an export of the entire site's content.

The go to design would be to add `rss_limit` or similar to your `_config.yml` file, defaulting to something like `10`. That feels lame to me. Why would someone want 11? 25? 5? Feels like we're abstracting a variable for the sense of not hard-coding a number, but I don't see why the number can't be hard-coded. 

What's a good number of posts for a feed? The 10 most recent? 5 most recent? Let's figure out that magic number and just run with it, rather than pushing the question onto users.

benbalter:
An attempt to fix #40, or at least start the conversation, by limiting feeds to 10 posts.

Fixes #40.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -25,7 +25,7 @@
     </author>
   {% endif %}
 
-  {% for post in site.posts %}
+  {% for post in site.posts | limit: 10 %}
     <entry>
       <title>{{ post.title | markdownify | strip_html | strip_newlines | xml_escape }}</title>
       <link href="{{ post.url | prepend: url_base }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />


parkr:
someone is going to want to make this configurable... but 10 seems reasonable.

benbalter:
A lot of @pathawks's arguments over in #40 were persuasive to me, but think it's easier to add an option than to remove one. If our experience parallels WordPress's, and 99% of users are fine with a hard limit of 10 posts, I think that's more than enough of a core user case for me to say lets leave it out. 

That said, if someone submits a pull request to make in an option with a reason other than "it should be an option", I don't know that we need to be dogmatic here. 

Thanks for the :thought_balloon: @parkr. Going to give this a spin in the wild, and see what type of feedback we get from users, one way or the other.

pathawks:
> it's easier to add an option than to remove one.

:+1:

PR #42, opened 2015-05-20
Use jekyll-last-modified-at for updated date

### LINKED ISSUE #26: Pull post update date from jekyll_last_modified

pathawks:
If installed, use [**jekyll-last-modified-at**](https://github.com/gjtorikian/jekyll-last-modified-at) for a post's updated date.

Fixes #26


parkr:
cool!

benbalter:
I really like the fact that it falls back to post date if last modified at isn't installed. 

Think this is solid and mergeable as is, but was hoping to get a 0.1.x into GitHub Pages so we can get some feedback from a bigger user base, so going to suggest we hold off (hopefully a day or two) on merging this (in the event there are more bug fixes we find), so this can make its way in to 0.2 (without having to get crazy with maintaining parallel development branches for such a small plugin).

pathawks:
:cool: Thanks for the update.

PR #43, opened 2015-05-21
Add xml:base attribute to the feed tag

### LINKED ISSUE #22: Replace URLs with full URLs in post.content

benbalter:
Per @pathawks over in https://github.com/jekyll/jekyll-feed/pull/36#issuecomment-104422305, and per the Atom spec:

>  Any element defined by this specification MAY have an xml:base
   attribute [W3C.REC-xmlbase-20010627].  When xml:base is used in an
   Atom Document, it serves the function described in section 5.1.1 of
   [RFC3986], establishing the base URI (or IRI) for resolving any
   relative references found within the effective scope of the xml:base
   attribute.

I believe this fixes #22.

/cc @holman and @bkeepers, as this should finally fix [the question that started this all](https://twitter.com/holman/status/597777659288887296)

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -4,12 +4,11 @@
 {% else %}
   {% assign url_base = site.github.url %}
 {% endif %}
-<feed xmlns="http://www.w3.org/2005/Atom">
+<feed xmlns="http://www.w3.org/2005/Atom" xml:base="{{ url_base | xml_escape }}">
   <generator uri="http://jekyllrb.com" version="{{ jekyll.version }}">Jekyll</generator>
   <link href="{{ page.url | prepend: url_base }}" rel="self" type="application/atom+xml" />
   <link href="{{ url_base }}/" rel="alternate" type="text/html" />
   <updated>{{ site.time | date_to_xmlschema }}</updated>
-  <xml:base>{{ url_base | xml_escape }}/</xml:base>
   <id>{{ url_base | xml_escape }}/</id>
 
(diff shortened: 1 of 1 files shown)


pathawks:
We might need to add it *per post*, if a post contains relative links.

benbalter:
> We might need to add it per post, if a post contains relative links.

Done via  d863e8a. Good call.

lib/feed.xml
```diff
@@ -33,7 +33,9 @@
       <updated>{{ post.date | date_to_xmlschema }}</updated>
 
       <id>{{ post.id | prepend: url_base | xml_escape }}</id>
-      <content type="html">{{ post.content | markdownify | xml_escape }}</content>
+      <content type="html" xml:base="{{ url_base | xml_escape }}">
```
pathawks: I think it should be `post.url` rather than just `url_base`, since that is what links would be relative to.

benbalter: Once again, you're absolutely right.

pathawks:
:+1:

benbalter:
[It works!](http://ben.balter.com/feed.xml)

pathawks:
:tada:

PR #45, opened 2015-05-22
Use uri in all places

benbalter:
For `site.author` we used `uri`, but for `post.author` we used `url`. I personally like `url` better, it's less pretentious, but went with `uri` in all cases, to match the documentation.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -59,8 +59,8 @@
           {% if post.author.email %}
             <email>{{ post.author.email | xml_escape }}</email>
           {% endif %}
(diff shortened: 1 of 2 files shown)


pathawks:
The [test is wrong](https://github.com/jekyll/jekyll-feed/blob/master/spec/jekyll-feed_spec.rb#L57); should be updated to look for `/<author>\s*<name>Ben</name>\s*<email>ben@example.com</email>\s*<uri>http://ben.balter.com</uri>\s*</author>/`

benbalter:
Good catch. Tests now pass.

PR #46, opened 2015-05-23
Allow posts to override post.id

pathawks:
This PR contains a failing test.

Apparently, it is not possible for a post to override it's Jekyll assigned `post.id`?

### DIFF
--- a/spec/jekyll-feed_spec.rb
+++ b/spec/jekyll-feed_spec.rb
@@ -85,6 +85,11 @@ describe(Jekyll::JekyllFeed) do
     expect(contents).not_to match /Liquid is not rendered\./
   end
 
+  it "allows posts to override the default post.id" do
+    expected = "<id>postID</id>"
+    expect(contents).to include(expected)
+  end
+
   context "parsing" do
     let(:feed) { RSS::Parser.parse(contents) }
 
@@ -105,9 +110,9 @@ describe(Jekyll::JekyllFeed) do
 
     it "includes the items" do
       if Gem::Version.new(Jekyll::VERSION) > Gem::Version.new('3')
-        expect(feed.items.count).to eql(8)
-      else
         expect(feed.items.count).to eql(9)
+      else
+        expect(feed.items.count).to eql(10)
       end
     end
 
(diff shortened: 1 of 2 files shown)


pathawks:
jekyll/jekyll#3725

parkr:
Not sure how I feel about this, but we are pretty forgiving when it comes to automagical behavior. I defer to @benbalter's judgement.

benbalter:
What's your use case for needing to explicitly specify a post ID? (you are correct, Jekyll doesn't let you override the post ID).

parkr:
> What's your use case for needing to explicitly specify a post ID?

@benbalter IIRC, it was to match historical documents from other systems, e.g. migrating from Tumblr to Jekyll but keeping the ID's the same for old posts.

benbalter:
> to match historical documents from other systems

I thought Jekyll ignored the `ID` in front matter, in favor of its own (for posts)? That may have changed with Jekyll 3.

pathawks:
> to match historical documents from other systems

The way feed readers keep track of which posts have been read is with the posts’ IDs.

When migrating to Jekyll from a different system, if all of the posts in the feed recieve new IDs, then all of the posts will appear unread to all clients of the feed. Keeping the original ID gets around this issue.

There should be *some* way for a post to let ‘jekyll-feed‘ know what that original ID is.

parkr:
> I thought Jekyll ignored the ID in front matter, in favor of its own (for posts)? That may have changed with Jekyll 3.

It does, and (I think) always did. `Document#id` and `Docuement#data['id']` would then yield different things in this case – one defined by the filename, one by the front matter.

pathawks:
Would it be acceptable for the sitemap to use `post.guid` instead of `post.id` if it is present?

parkr:
Question: how many users of jekyll-feed have posts with an `id` set whose feeds would be broken as a result of this new feature?

parkr:
Is this the case in Jekyll 3.0.3 and 3.1.2?

spec/jekyll-feed_spec.rb
```diff
@@ -125,9 +130,9 @@
 
     it "doesn't include the item's excerpt if blank" do
       if Gem::Version.new(Jekyll::VERSION) > Gem::Version.new('3')
-        post = feed.items.first
```
pathawks: This is a very fragile test. Will need to be updated anytime a new post is added to the feed or an old post is updated, since posts are sorted by updated date. :unamused:

pathawks:
@parkr [Still failing](https://travis-ci.org/jekyll/jekyll-feed/builds/117364411#L255) on 3.1.2

pathawks:
As nice as it would be to preserve all the `ID`s of posts when migrating to Jekyll and **jekyll-feed**, it's not going to happen _this_ way. 😞

PR #49, opened 2015-05-24
Fix Liquid error

### LINKED ISSUE #48: Liquid syntax error in jekyll 3.0
`jekyll 3.0.0.pre.beta5`

While trying to serve:

`  Liquid Exception: Liquid syntax error: Expected end_of_string but found pipe in "post in site.posts | limit: 10" in feed.xml`

Not sure why liquid is processing `feed.xml`, I thought anything not prefixed with `_` won't get processed.

pathawks:
According to the [Liquid docs](https://github.com/Shopify/liquid/wiki/Liquid-for-Designers#for-loops), we should not include a pipe before `limit`.

Fixes #48

We should also add a test that makes sure we are actually limiting to 10 posts. Right now, there just aren't that many posts in our test site. Should that be part of this PR, or another?

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -35,7 +35,7 @@
     </author>
   {% endif %}
 
-  {% for post in site.posts | limit: 10 %}
+  {% for post in site.posts limit: 10 %}
     <entry>
       <title>{{ post.title | markdownify | strip_html | strip_newlines | xml_escape }}</title>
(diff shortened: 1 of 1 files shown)


benbalter:
Huh. Never knew that. Odd that it works. Thanks!

rayshan:
Thanks guys

PR #53, opened 2015-06-02
Loosen Jekyll Version Requirements.

envygeeks:
Currently we default install jekyll-feed across all our Docker images (master,beta included) and because of the tight coupling that happens on versioning we end up with multiple versions of Jekyll, this loosens that requirement since I don't assume this would break but if it does during this pull I'll gladly fix any problems so it works with both 2.5 and 3.0.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,8 +13,13 @@ matrix:
   exclude:
     - rvm: 1.9
       env: JEKYLL_VERSION=3.0.0.beta5
+    - env: JEKYLL_VERSION=2.4
+      rvm: 2.1
+    - rvm: 2.2
+      env: JEKYLL_VERSION=2.4
 
 env:
   matrix:
     - ""
+    - JEKYLL_VERSION=2.4
     - JEKYLL_VERSION=3.0.0.beta5
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,8 +3,18 @@ rvm:
 - 2.1
 - 2.2
 - 2.0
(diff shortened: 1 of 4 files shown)


parkr:
:+1:

benbalter:
Can you make it >= 2.4 (what Pages currently runs)? Otherwise :+1:. Nice implementation. Thanks for the pull request.

envygeeks:
@benbalter yup! Once I confirm that the current build works (since I added 2.2 and testing on 3.0 beta -- I don't know if you seen that) I'll drop it to 2.4 and add that to the matrix (unless you want to skip that in the matrix?)

.travis.yml
```diff
@@ -1,9 +1,14 @@
 language: ruby
 rvm:
 - 2.1
-- 2.0.0
+- 2.2
+- 2.0
 - 1.9.3
```
parkr: Jekyll 3 won't work here :( :(

envygeeks: Yeah I'm trying to work out the matrix exclude so I can skip this test on 1.9.3.

envygeeks:
I have completed the matrix and figured that beast mode travis-ci configuration out.  Now jekyll-feed tests 2.4 - 3.0 on select Ruby versions.  This is safe for review and merge when everybody has :+1:

parkr:
:shipit:

benbalter:
Nice. Thanks again @envygeeks. :turtle:

PR #54, opened 2015-06-03
Custom feed path setting

orderedlist:
Many sites already have feeds setup at a path other than `feed.xml`. If their host e.g., GitHub pages, doesn’t allow for redirects, changing to a new feed path isn’t really an option.

This PR adds a `_config.yml` option of `feed_path`, which defaults to `feed.xml` but can be set to a custom path like `atom.xml`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,21 @@ The plugin will automatically use any of the following configuration variables, 
 * `name` - The title of the site, e.g., "My awesome site"
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
 * `url` - The URL to your site, e.g., `http://example.com`. If none is provided, the plugin will try to use `site.github.url`.
-* `feed_path` - The path to the generated feed, e.g., `atom.xml`. If none is provided, the plugin will default to `feed.xml`.
 * `author` - Your name, e.g., "Dr. Jekyll." This can be a string (with the author's name), or an object with the following properties:
   - `name` - **Required** Display name of the author
   - `email` - Email address of the author
   - `uri` - Webpage where more information about the author can be found
 
+### Already have a feed path?
+
+Do you already have an existing feed someplace other than `/feed.xml`, but are on a host like GitHub Pages that doesn't support machine-friendly redirects? If you simply swap out `jekyll-feed` for your existing template, your existing subscribers won't continue to get updates. Instead, you can specify a non-default path via your site's config.
+
+```yml
+feed:
+  path: atom.xml
+```
+
+To note, you shouldn't have to do this unless you already have a feed you're using, and you can't or wish not to redirect existing subscribers.
 
 ### Optional front matter
 
--- a/README.md
+++ b/README.md
@@ -30,6 +30,7 @@ The plugin will automatically use any of the following configuration variables, 
 * `name` - The title of the site, e.g., "My awesome site"
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
 * `url` - The URL to your site, e.g., `http://example.com`. If none is provided, the plugin will try to use `site.github.url`.
+* `feed_path` - The path to the generated feed, e.g., `atom.xml`. If none is provided, the plugin will default to `feed.xml`.
 * `author` - Your name, e.g., "Dr. Jekyll." This can be a string (with the author's name), or an object with the following properties:
   - `name` - **Required** Display name of the author
   - `email` - Email address of the author
--- a/lib/jekyll-feed.rb
+++ b/lib/jekyll-feed.rb
@@ -12,6 +12,10 @@ module Jekyll
       @context.registers[:site].config
     end
 
+    def path
+      config["feed_path"] || "feed.xml"
+    end
+
     def url
       if config["url"]
         config["url"]
@@ -22,7 +26,7 @@ module Jekyll
 
     def render(context)
       @context = context
-      "<link type=\"application/atom+xml\" rel=\"alternate\" href=\"#{url}/feed.xml\" title=\"#{config["name"]}\" />"
+      "<link type=\"application/atom+xml\" rel=\"alternate\" href=\"#{url}/#{path}\" title=\"#{config["name"]}\" />"
     end
   end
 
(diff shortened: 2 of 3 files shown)


lib/jekyll-feed.rb
```diff
@@ -12,6 +12,10 @@ def config
       @context.registers[:site].config
     end
 
+    def path
+      config["feed_path"] || "feed.xml"
```
benbalter: We'll need to use `Jekyll.sanitized_path(@site.source, config["feed_path"])` here. A malicious user could provide a relative path outside the Jekyll site source to expose the contents of files they otherwise wouldn't have access to. It may also be worth adding a test for this (e.g., verify a path of `../../foo` becomes `/foo`).

README.md
```diff
@@ -30,6 +30,7 @@ The plugin will automatically use any of the following configuration variables,
 * `name` - The title of the site, e.g., "My awesome site"
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
 * `url` - The URL to your site, e.g., `http://example.com`. If none is provided, the plugin will try to use `site.github.url`.
+* `feed_path` - The path to the generated feed, e.g., `atom.xml`. If none is provided, the plugin will default to `feed.xml`.
```
benbalter: What do you think about:

```yml
feed:
  path: `atom.xml`
```

This will create a feed namespace (similar to other plugins), in case we have other options in the future.

I may also encourage the default a bit more. Reading it now, I feel like I have to make a decision about what the "best' feed path is. I'd make it clear that the only time you might use this would be if you had an existing feed, and are using GItHub pages, and don't want to break existing subscribers.

How about a separate section with something like

> # Already have a feed path?
>
> Do you already have an existing feed someplace other than `/feed.xml`, but are on a host like GitHub Pages that doesn't support machine-friendly redirects? If you simply swap out `jekyll-feed` for your existing template, your existing subscribers won't continue to get updates. Instead, you can specify a non-default path via your site's config.

```yml
feed:
  path: atom.xml
```

To note, you shouldn't have to do this unless you already have a feed you're using, and you can't redirect existing subscribers.

benbalter:
Thanks for the pull request @orderedlist. Small doc suggestion, and small security fix, otherwise, this :eyes: good.

parkr:
:)

README.md
orderedlist: I think that makes sense. :+1:

lib/jekyll-feed.rb
orderedlist: Is this actually the case? This `path` is only used as a string in the `link` tag rendered into the pages, it doesn't read the file at all.

benbalter: Ahh. I was thinking the source template. We sanitize prior to the output destination (to prevent overwriting a file outside the site source). 

@parkr is there any problem with it being passed into `keep_files`? Should we sanitize there to be safe?

lib/jekyll-feed.rb
```diff
@@ -12,6 +12,14 @@ def config
       @context.registers[:site].config
     end
 
+    def path
+      if config["feed"] && config["feed"]["path"]
```
aroben: Will this blow up if someone does `feed: foo` in their config file? In that case `config["feed"]` won't be a dict.

benbalter: Goot catch. Here's [how we did it for jemoji](https://github.com/jekyll/jemoji/blob/master/lib/jemoji.rb#L14-L21):

```ruby
 def src
      @src ||=
        if config.key?("emoji") && config["emoji"].key?("src")
          config["emoji"]["src"]
        else
          "https://assets.github.com/images/icons/"
        end
    end
```

aroben: Won't `key?` have the same problem? There's no `String#key?` method.

parkr: perhaps

```ruby
def path
  if config.key?("feed")
    userpath = config["feed"].is_a? Hash
      config["feed"]["path"]
    else
      config["feed"].to_s
    end
    Jekyll.sanitized_path(site.source, userpath)
  end
end
```

envygeeks: The only thing that matters here is if this is supposed to be a hash, if it is then let it blow up because that's your fault and it's the right thing to do, if accepts any value then this matters.

envygeeks:
This is broken on 3.0, is this going to be fixed because we have a lot of users on 3.0 and this is a defacto-plugin since it sits under the Jekyll umbrella.

benbalter:
> This is broken on 3.0

Can you explain what you mean by that?

envygeeks:
@benbalter https://travis-ci.org/jekyll/jekyll-feed/jobs/65270484

envygeeks:
Actually, this failing test confuses me based on the way it's failing, I'm going to patch a local version and drop myself into a debugger and check out the state machine across all Jekyll versions because logically if it failed on one it should fail on all no?

lib/jekyll-feed.rb
aroben: Yeah, if the goal is only to support `feed` as a hash then blowing up might be a good thing. I don't know the conventions around this kind of thing in Jekyll-land which is why I raised it.

envygeeks: It's not up to Jekyll-Core it's up to the author of the plugin or in this case @orderedlist, if he wants a hash then that's on him tbh, Jekyll doesn't enforce like that (not that I've seen at least,) so it's entirely up to the author if he wants to blow up because he wants a hash or not, it's not up to Jekyll-Core whether he accepts both gracefully.

orderedlist: So the way it is now:

```rb
if config["feed"] && config["feed"]["path"]
```

...won't blow up unless feed.path is set and incorrect (e.g. a hash instead of a string). I'm personally fine with this, as it's well documented how it should properly be configured.

envygeeks:
I finally had a second to look at the tests to figure out what was going on and it turns out that the site wasn't regenerating on Jekyll 3 so it wasn't picking up what @orderedlist wanted but I also had to make another small change to the source to get the test to pass:

```ruby
    let(:config) do
      Jekyll.configuration(Jekyll::Utils.deep_merge_hashes(
        overrides, {
          "feed" => {
            "path" => "atom.xml"
          }
        }
      ))
    end
```

That way the new config would pick up the original overrides with regeneration.

orderedlist:
I think I've addressed all the feedback. :space_invader:

benbalter:
Is there any issue with adding an unsanitized, potentially relative or synmlink path to keep files? @parkr?

parkr:
The path is sanitized later, I thought?

benbalter:
> The path is sanitized later, I thought?

`keep_files` can already accept arbitrary input via `_config.yml`, so there's no additional risk of arbitrary input here. 

@orderedlist thanks for submitting the PR and for working through the feedback. WIll try to get a new release out so you can test it out in the wild. :smile_cat:

orderedlist:
yes

envygeeks:
:heart:

parkr:
@orderedlist that gif tho. :joy: 

:+1:

Issue #66, opened 2015-09-19
Handling of whitespace in title / excerpt


maul-esel:
## Problem
The current `feed.xml` strips newlines from the title and excerpt. However, that leaves no space between the last word of a line and the first word of the next line.

## Example
The markdown
```
This is some
text
```

leads to `This is sometext`.

## Suggestion
Replace `strip_newlines` with `replace: '\n', ' '`.

**Update:** or rather `replace: '\r\n', ' ' | replace: '\n', ' '`.

benbalter:
> Replace `strip_newlines` with `replace: '\n', ' '`.

Is there an example of when we wouldn't want `strip_newlines` to behave that way?

maul-esel:
I'm not sure I understand the question. Are you suggesting to make it the default behaviour of `strip_newlines` to insert spaces?

benbalter:
> Are you suggesting to make it the default behaviour of `strip_newlines` to insert spaces?

Yep. Can you think of when a user *wouldn't* want that behavior? I assumed that's how it worked.

maul-esel:
I can't think of one, but who knows...

Also, overriding `strip_newlines` *in jekyll* would cause the same Liquid markup used with jekyll to behave differently than when used somewhere else – that might be quite confusing. The alternative would be to try patching it upstream.

pathawks:
I would call this a `strip_newlines` issue.

maul-esel:
I disagree. `strip_newlines` is supposed to remove newlines, and that's exactly what it does, nothing more and nothing less.

parkr:
We'd want a `replace: '\n', ' '` situation.

zhangzheng08331:
Hello!Please give an example if we wouldn't want strip_newlines to behave that way?

pathawks:
This should be fixed by #67

PR #67, opened 2015-09-20
Replace newlines with space

pathawks:
Currently failing tests because this method does not remove trailing space from the end of the title if the title ends with a newline.

Not sure what to do about that without chasing our own tail. Maybe we just don't strip newlines at all?

#66


parkr:
Is there a "strip" filter we could use?

pathawks:
Shopify has one, but Jekyll does not.

parkr:
> Shopify has one, but Jekyll does not.

¡ABSURD! Is it in Liquid 3 only, maybe?

pathawks:
[Looks that way](https://github.com/Shopify/liquid/commit/d2827bfa76801d0e8b104ab28703ce86164d5253)

mattr-:
welp. time to upgrade :troll:

parkr:
Will release a new beta. Maybe I can get some time this week...

pathawks:
Using the `strip` works for Jekyll 3.

Jekyll 2 will not have trailing whitespace removed, but calling a filter that doesn't exist doesn't break anything either with the relaxed Liquid parsing.

pathawks:
The other option would be to just add our own Liquid filter to collapse whitespace

```ruby
  def collapse_whitespace(input)
    input.to_s.gsub(/\s+/, ' ').strip
  end
```

:question:

parkr:
Any liquid filters we make here will be globally available, just a head's up.

pathawks:
> Any liquid filters we make here will be globally available, just a head's up.

Is that to be avoided?

parkr:
@pathawks Would you please add a test for this?

pathawks:
Test added :+1:

PR #68, opened 2015-09-22
Remove unused files from test

pathawks:
These were all left over from `jekyll-sitemap`; we were not using any of them.

I did remove _one_ post that was not being used for anything, and this is what affected the number of posts expected in the output feed.


parkr:
Thanks, @pathawks !!!

Issue #15, opened 2015-05-12
Allow specifying the language code


XhmikosR:
By default, don't add anything, but give the user the ability to specify the language code.

pathawks:
Is there any prior art here?

Do any of the importers import any language information, either per post or per site?

Is there any existing convention for specifying this information?

XhmikosR:
Only the specs as far as I can tell. `language` is a per site element.

pathawks:
But, because it XML, any element can have an `xml:lang` attribute.

XhmikosR:
That is true, but the specs don't mention anything about each element. So it should be a per channel sub-element.

pathawks:
Is there any prior art here at all? Are there any **i18n** Jekyll plugins/workflows that set language information anywhere?

parkr:
> Are there any i18n Jekyll plugins/workflows that set language information anywhere?

@pathawks There are a number of i18n plugins for Jekyll out in the wild. Octopress's Multilingual is the only one whose development I have followed. https://github.com/octopress/multilingual

pathawks:
I don't see any `language` element in the Atom spec, but `<feed>` could have an `xml:lang` set to `{{ site.lang }}`, if present. Each `<entry>` could also have an `xml:lang` set to `{{ post.lang }}`. All links could also include `hreflang` if we can reasonably figure out the language of the resource the link points to.

Would this be reasonable/sufficient? Are there any great examples of feeds from multilingual sites that properly include language information to look to as an example?

pathawks:
This has been fixed by #75

XhmikosR:
Thanks! Any chance for a new release?

PR #75, opened 2015-12-13
Think about how i18n might work

pathawks:
This pull request represents an implementation of my proposal in [#15 (comment)](https://github.com/jekyll/jekyll-feed/issues/15#issuecomment-164300135)

Certainly, some tests need to be written before this is merged, but this is an effort to get the discussion going.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -4,10 +4,10 @@
 {% else %}
   {% assign url_base = site.github.url %}
 {% endif %}
-<feed xmlns="http://www.w3.org/2005/Atom">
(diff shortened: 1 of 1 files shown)


parkr:
This seems reasonable to me, as it is congruent with the idea that the entire site is one language, or that a given post is one language. @benbalter, how do you feel? It seems like a small thing, but could make a number of folks happy.

benbalter:
:+1:. Is `lang` an established property? (versus `language:`?)

pathawks:
`lang` is [used by **octopress/multilingual**](https://github.com/octopress/multilingual#setup)

benbalter:
Makes sense.

parkr:
:shipit:

pathawks:
@jekyllbot: merge

PR #80, opened 2016-01-05
feed_meta Should include baseurl

### LINKED ISSUE #79: Incorrect `link[href]` value if baseurl is provided in _config.yml
Given the following `_config.yml` file content:

```yml
baseurl: /blog
feed:
  path: feed.xml
```

The liquid tag `{% feed_meta %}` renders:

```
<link type="application/atom+xml" rel="alternate" href="feed.xml" title="..." />
```

(instead of)

```
<link type="application/atom+xml" rel="alternate" href="/blog/feed.xml" title="..." />
```

Whereas the feed is properly generated at `localhost:4000/blog/feed.xml` (and not at localhost:4000/feed.xml`.

If I change the value of the `path` to `blog/feed.xml`, the `link[href]` value is correct but then the feed content is not rendered at the given URL.

Is there anything special to do? I thought we'd have followed the README correctly.

Thanks :-)

refs sudweb/blog#10

pathawks:
Fixes #79

### DIFF
--- a/lib/jekyll-feed.rb
+++ b/lib/jekyll-feed.rb
@@ -28,7 +28,7 @@ module Jekyll
 
     def url
       if config["url"]
-        config["url"]
+        URI.join(config["url"], config["baseurl"])
       elsif config["github"] && config["github"]["url"]
         config["github"]["url"]
       end
--- a/spec/jekyll-feed_spec.rb
+++ b/spec/jekyll-feed_spec.rb
@@ -182,6 +182,12 @@ describe(Jekyll::JekyllFeed) do
       expect(contents).to match /http:\/\/example\.org\/bass\/2014\/03\/02\/march-the-second\.html/
       expect(contents).to match /http:\/\/example\.org\/bass\/2013\/12\/12\/dec-the-second\.html/
     end
+
+    it "renders the feed meta" do
+      index = File.read(dest_dir("index.html"))
+      expected = '<link type="application/atom+xml" rel="alternate" href="http://example.org/bass/feed.xml" title="My awesome site" />'
+      expect(index).to include(expected)
+    end
   end
 
   context "feed meta" do


lib/jekyll-feed.rb
```diff
@@ -28,7 +28,7 @@ def path
 
     def url
       if config["url"]
-        config["url"]
+        config["url"] + config["baseurl"]
```
benbalter: Would it be more robust to use URI.join here?

pathawks: Good call :+1:

benbalter:
@pathawks did we lose the test when you force pushed from a375d8b to b56c3a3? Last, do we need to test with and without a base url to ensure, e.g., we don't get double trailing slashes?

pathawks:
> @pathawks did we lose the test

Eek! Sorry! Thanks for catching that

> do we need to test with and without a base url

We already have [a test without base url](https://github.com/jekyll/jekyll-feed/blob/v0.4.0/spec/jekyll-feed_spec.rb#L187-L193). Perhaps I should add this test to the `"feed meta"` context instead of the `"with a baseurl"` context? How can we organize this to make the most sense?

benbalter:
> We already have a test without base url. Perhaps I should add this test to the "feed meta" context instead of the "with a baseurl" context? How can we organize this to make the most sense?

Nah, just misread the diff. This is awesome. Thanks for tracking this down and putting this together. :+1:

pathawks:
This might break some things.

Looking at [the source](https://github.com/jekyll/jekyll-feed/blob/v0.4.0/lib/jekyll-feed.rb#L21-L40), if somebody specifies a custom path for the feed, it does *not* currently respect `baseurl`. I believe it *should*, but we will have to make sure to let folks know that if the are upgrading and have a custom `feed.path` *and* `baseurl`, they will need to update the `feed.path`

PR #82, opened 2016-01-06
Consolidate regexs

pathawks:
Inspired by @benbalter, I have consolidated three regular expressions into one.


parkr:
Are each of these tested?

pathawks:
The whole point of all of these was to strip unnecessary whitespace, which [there is a test for](https://github.com/jekyll/jekyll-feed/blob/master/spec/jekyll-feed_spec.rb#L35-L38).

parkr:
Excellent.

parkr:
Thanks! :tada:

PR #83, opened 2016-01-11
Use Module#method_defined?

envygeeks:
I don't think the old way works the way you think it worked.


envygeeks:
Apparently it does.

pathawks:
:+1:

parkr:
@jekyllbot: merge +minor

PR #86, opened 2016-01-19
Separate each class/module into its own file

pathawks:
There is too much going on in this plugin to keep everything in one file with nearly 100 lines. Let's break things up a bit.

Am I namespacing everything right? Everything was a child of the `Jekyll` module, so that's why I put it all in a `jekyll` subdirectory.


lib/jekyll-feed.rb
```diff
@@ -1,94 +1,4 @@
...
-    end
-  end
-end
-
-unless Liquid::StandardFilters.method_defined?(:strip)
-  Liquid::Template.register_filter(Jekyll::StripWhitespace)
-end
-
-Liquid::Template.register_tag("feed_meta", Jekyll::FeedMetaTag)
+require "jekyll/feedmetatag"
```
pathawks: Would it make more sense to move this to `jekyll/tags/feedmeta`?

pathawks:
7d7d003 Only added a comment; no reason it should have broken the build.

Is [this](https://travis-ci.org/jekyll/jekyll-feed/jobs/103250449#L187) just a flaky test.

pathawks:
Why is that test failing??? :rage2:

lib/jekyll/feedmetatag.rb
```diff
@@ -0,0 +1,32 @@
+module Jekyll
+  class FeedMetaTag < Liquid::Tag
```
benbalter: I've always seen [files named with snake_case](https://github.com/bbatsov/ruby-style-guide#snake-case-files), meaning the file should be named `feed_meta_tag`.

lib/jekyll/jekyllfeed.rb
```diff
@@ -0,0 +1,50 @@
+require "fileutils"
```
benbalter: This file name feels strange to me, since we also have a `lib/jekyll-feed.rb` as the primary file.

Technically, I believe it should be `lib/jekyll/feed.rb`, with sub-files in `lib/jekyll/feed`, but then also technically, the gem should be required as `jekyll/feed`, however, that's not the design pattern for any of the other gems that I've seen.

Most use `lib/jekyll-feed.rb` as the primary file, with `lib/jekyll-feed/foo.rb` as sub files, which isn't 100% in line with Ruby Gem conventions, but allows us to do `gems: [jekyll-feed]`, and not `gems: [jekyll/feed]`. 

@parkr any strong thoughts one way or the other here?

lib/jekyll/pagewithoutafile.rb
```diff
@@ -0,0 +1,7 @@
+module Jekyll
```
benbalter: Regardless of the above, this should remain in `lib/jekyll`, as it's an extension of Jekyll, not jekyll-feed.

benbalter:
:+1: on the idea. A few specifics to work out in terms of establishing a naming convention for plugins, that I've been struggling with myself.

pathawks:
@benbalter How does this look?

benbalter:
@pathawks looks :sparkles:. Thanks (and thanks for working through it with me!).

pathawks:
Thanks for the help :+1:

PR #87, opened 2016-01-20
Do not let jekyll-last-modified-at break tests

pathawks:
I don't know why #86 is failing tests on Travis, but I cannot reproduce locally.

Frankly, I don't much care if **jekyll-last-modified-at** is coming up with the correct date or not, so long as we are using what is provided in the `last_modified_at` property.

This removes the dev dependency on **jekyll-last-modified-at** and, hopefully, will make the tests a little less rickety…?


benbalter:
Makes sense. Think we can assume the plugin has its own tests, and just use its documented API.

Issue #94, opened 2016-02-08
Exclude Drafts from feed


jeremiemartinez:
Hi there,

What do you think about filtering future posts from the `feed.xml` ?

To give you some context, most of the time I write a new article, I publish it with a future date, it is not listed and I send direct link to friends for proof reading. My problem is that even if my article has a future date, it still appears in the `feed.xml`...

Would you be willing to accept a PR if I work on it ? Is there any other way to filter posts appearing in the `feed.xml` ?

Thanks,

pathawks:
I have also published a post “publically” for the purpose of sending it to somebody to proofread.

I’m not sure if future date is the way to do this, seems like that could lead to some trouble if time zones or something caused Jekyll to improperly exclude a new post from the feed.

Perhaps a way to explicitly exclude a post from the feed, similar to `sitemap: false`?

jeremiemartinez:
So we could have something like: `feed: false` that could be added to the frontmatter of each post and then filtered in the for loop in `feed.xml`. Is that what you mean ? It would also be a good solution since I understand your concerns about time zones.

pathawks:
> So we could have something like: `feed: false` that could be added to the frontmatter of each post and then filtered in the for loop in `feed.xml`. Is that what you mean ?

:+1:

benbalter:
@jeremiemartinez how do you prevent future posts from appearing in your site's human-readable index? 

I'm wondering if rather than have the user specify the feature they want ("don't include this in the sitemap", "don't include this in the feed"), the user should expose intend ("this shouldn't be public"), and the two plugins should respond accordingly.

jeremiemartinez:
@benbalter Well, my approach is very similar that the one advised by @pathawks. However, don't judge me, this is not the best code I wrote...

First, I add `future: true` to my `_config.yml`. 
Then, I declare in the frontmatter: `public: 1`
Finally, in my `index.html`, I loop and filter:
```
{% for post in paginator.posts %}
   {% if post.public != 1  %}
        […]
   {% endif %}
 {% endfor %}
```

Your idea about a public flag that could be used for every plugin is great but it would require to directly modify jekyll to expose such an API I guess :/

pathawks:
Jekyll has a **Drafts**  feature already: is there any way in Liquid to detect if a post is a draft?

[...]

PR #95, opened 2016-02-09
Filter drafts from feed.xml

jeremiemartinez:
Fixing #94

### DIFF
--- a/spec/jekyll-feed_spec.rb
+++ b/spec/jekyll-feed_spec.rb
@@ -7,6 +7,7 @@ describe(Jekyll::JekyllFeed) do
       "full_rebuild" => true,
       "source"      => source_dir,
       "destination" => dest_dir,
+      "show_drafts" => true,
       "url"         => "http://example.org",
       "name"       => "My awesome site",
       "author"      => {
@@ -41,6 +42,9 @@ describe(Jekyll::JekyllFeed) do
     expect(contents).to match /http:\/\/example\.org\/2014\/03\/04\/march-the-fourth\.html/
     expect(contents).to match /http:\/\/example\.org\/2014\/03\/02\/march-the-second\.html/
     expect(contents).to match /http:\/\/example\.org\/2013\/12\/12\/dec-the-second\.html/
+    if Gem::Version.new(Jekyll::VERSION) > Gem::Version.new('3')
+      expect(contents).to_not match /http:\/\/example\.org\/2016\/02\/09\/a-draft\.html/
+    end
   end
 
   it "does not include assets or any static files that aren't .html" do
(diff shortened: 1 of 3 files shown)


pathawks:
We will need to add a test for this: a drwft post that is generated but excluded from the feed.

jeremiemartinez:
Sure. I was thinking about it actually. I am pretty new to Ruby but I will look into it tonight.

jeremiemartinez:
Hmmm, looks like this is only working with Jekyll 3.0 :(

jeremiemartinez:
Do you have any idea about what could be a solution for this ?

benbalter:
I'm fine with this being a Jekyll 3-only feature. If so, you'd need to wrap the test in `if Jekyll::VERSION > "3"`.

spec/jekyll-feed_spec.rb
```diff
@@ -116,7 +124,11 @@
     end
 
     it "doesn't include the item's excerpt if blank" do
-      post = feed.items.first
+      if Gem::Version.new(Jekyll::VERSION) > Gem::Version.new('3')
```
jeremiemartinez: Before Jekyll 3.0, it looks like draft were generated with today date path, therefore leading first item was the draft instead of the real first post.

spec/jekyll-feed_spec.rb
```diff
@@ -100,7 +104,11 @@
     end
 
     it "includes the items" do
-      expect(feed.items.count).to eql(8)
+      if Gem::Version.new(Jekyll::VERSION) > Gem::Version.new('3')
+        expect(feed.items.count).to eql(8)
+      else
+        expect(feed.items.count).to eql(9)
```
jeremiemartinez: new draft counts in post list before Jekyll 3.0

spec/jekyll-feed_spec.rb
pathawks: Nice sleuthing! :+1:

jeremiemartinez:
Let me know if you need more to merge this.

benbalter:
@parkr any strong feelings one way or the other here?

parkr:
> @parkr any strong feelings one way or the other here?

I have no _strong_ feelings, but I feel drafts should be in the feed if they're in `site.posts`. That said, I also feel folks shouldn't publish a site with `show_drafts: true` in production. This PR at least protects against publishing drafts by accident.

:+1:

benbalter:
Nice work @jeremiemartinez! Thanks! :cake: :tangerine: :necktie:

jeremiemartinez:
Thanks! That's great news :)

What is your release policy ? Do you plan to do one soon?

PR #99, opened 2016-03-03
Only test against Jekyll 3

pathawks:
This reduces the number of CI test environments from 7 to 1.

Since most of the business of this plugin is implemented in Liquid, *hopefully* we're not doing anything that might break in one version of Ruby and not another.

**This also drops tests against Jekyll 2.x**

I have not removed Jekyll 2.x from the `.gemspec`, I have only removed the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,19 +1,9 @@
 language: ruby
 rvm:
   - 2.2
-  - 2.1
-  - 2.0
-matrix:
-  include:
-    - # Ruby 1.9
-      rvm: 1.9
-      env: JEKYLL_VERSION=2.0
 env:
   global:
     - NOKOGIRI_USE_SYSTEM_LIBRARIES=true
-  matrix:
-    - JEKYLL_VERSION=3.0
-    - JEKYLL_VERSION=2.5
 cache: bundler
 sudo: false
 before_script: bundle update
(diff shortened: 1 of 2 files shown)


benbalter:
Should we [update the Gemspec](https://github.com/jekyll/jekyll-feed/blob/master/jekyll-feed.gemspec#L17) to require Jekyll > 3?

pathawks:
> Should we update the Gemspec to require Jekyll > 3?

I thought about that.

Since we have nothing to immediately gain by dropping support for 2.x, maybe we could hold off on this until GitHub Pages supports 3.1. At *that* time, we could drop support for anything less than 3.1 and take advantage of the `smartify` filter?

benbalter:
> Since we have nothing to immediately gain by dropping support for 2.x, maybe we could hold off on this until GitHub Pages supports 3.1. At that time, we could drop support for anything less than 3.1 and take advantage of the smartify filter?

Makes sense.

parkr:
@pathawks feel free to merge. you should be able to use jekyllbot too! If you want categories for the changelog, it understands:

- `+dev` – Development Fixes
- `+bug` – Bug Fixes
- `+major` – Major Enhancements
- `+minor` – Minor Enhancements
- `+site` – Site Enhancements

you can leave off the category to just add under the version. it uses the PR title so fix that too if you want to before you ask jekyllbot to merge. :smile:

parkr:
@jekyllbot: merge

PR #100, opened 2016-03-20
Use site.title for meta tag if available

pathawks:
This will use `site.title` (if present) for the feed title when rendering the `<link>` tag. If `site.title` is undefined, it will fallback to `site.name`.

This mirrors the current behavior of the [feed itself](https://github.com/jekyll/jekyll-feed/blob/v0.4.0/lib/feed.xml#L14-L18):

```html
  {% if site.title %}
    <title>{{ site.title | xml_escape }}</title>
  {% elsif site.name %}
    <title>{{ site.name | xml_escape }}</title>
  {% endif %}
```

### DIFF
--- a/lib/jekyll/feed_meta_tag.rb
+++ b/lib/jekyll/feed_meta_tag.rb
@@ -2,7 +2,7 @@ module Jekyll
   class FeedMetaTag < Liquid::Tag
     def render(context)
       @context = context
-      %(<link type="application/atom+xml" rel="alternate" href="#{url}/#{path}" title="#{config["name"]}" />)
+      %(<link type="application/atom+xml" rel="alternate" href="#{url}/#{path}" title="#{title}" />)
     end
 
     private
@@ -25,6 +25,10 @@ module Jekyll
       elsif config["github"] && config["github"]["url"]
         config["github"]["url"]
       end
+    end
+
+    def title
+      config["title"] || config["name"]
     end
   end
 end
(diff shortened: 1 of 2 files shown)


lib/jekyll/feed_meta_tag.rb
```diff
@@ -26,6 +26,14 @@ def url
         config["github"]["url"]
       end
     end
+
+    def title
```
benbalter: Perhaps 

```ruby
def title
  config["title"] || config["name"]
end
```

pathawks: I am trying to avoid output of `title=""` if neither is specified.

Am I overthinking this?

lib/jekyll/feed_meta_tag.rb
```diff
@@ -2,7 +2,15 @@ module Jekyll
...
     def render(context)
       @context = context
-      %(<link type="application/atom+xml" rel="alternate" href="#{url}/#{path}" title="#{config["name"]}" />)
+      link_attributes = [
+        'type="application/atom+xml"',
+        'rel="alternate"',
+        %(href="#{url}/#{path}")
+      ]
+      if title
+        link_attributes << %(title="#{title}")
```
pathawks: Is this too complicated for what I'm trying to accomplish?

benbalter: Not sure if it makes sense here, but [here's what we did for Jekyll avatar](https://github.com/benbalter/jekyll-avatar/blob/master/lib/jekyll-avatar.rb#L14-L33) to solve a similar problem.

pathawks:
Oh hey :100:

benbalter:
I think this is a lot cleaner (even if slightly over-engineered). 👍

pathawks:
@jekyllbot: merge +minor

PR #105, opened 2016-03-21
Drop support for Jekyll 2

pathawks:
As an addendum to #99, this drops test support for Jekyll 2 and bumps the required version of Jekyll.


pathawks:
@jekyllbot: merge

PR #106, opened 2016-03-21
Find author by reference

### LINKED ISSUE #62: Find author by reference
On my blog there are two authors. Their properties are defined in the `_config.yml`:

```yml
authors:
  foo:
    name: "Foo"
    email: "foo@domain.com"
    uri: "http://foo.com"
  bar:
    name: "Bar"
    email: "bar@domain.com"
    uri: "http://bar.com"
```

In the post's YAML frontmatter I put:

```yml
title: My Post
author: foo
```

Then later I extract the actual author with `site.authors[post.author]`. I do this so I don't have to write all those details in each post.

Is there a way to make something like this work with jekyll-feed?

pathawks:
Allow looking up author in `_data/authors.yml` in a way similar to **jekyll-seo-tag**

Fixes #62


benbalter:
Good call. (Although I fear a `jekyll-global-authors` or similar gem in our near future :smile:)

pathawks:
@jekyllbot merge

pathawks:
@jekyllbot: merge

Issue #56, opened 2015-06-04
Configure number of entries?


julien51:
Would it be possible to add an option for more than 10 entries?

benbalter:
@julien51 can you explain your use case and reasoning a bit more?

julien51:
Well, for high frequency feeds, 10 entries might be a little short, because the polling interval may be a bit longer than the time the 10 entries are "in the feed". A marge window, like 20 or more entries would be helpful.

10 is a good default though, just adding a setting to increase this would be nice!

benbalter:
Have you run into an issue with a site updating more frequently than the polling rate (or is this a hypothetical)? Would a default of 20 solve that? 15?

julien51:
Yes, I did run into such an issue (not with this generator though). The
feed updated roughly every 10 minutes and it was not rare to see pollers
fetch the feed once every 2 or 3 hours, resulting in missing updates.

I think 10 is a better default than 20 or 15, so don't change the
default... make this configurable for the few people (like me) for who this
could be a problem!
Thanks,

lancecarlson:
Perhaps this would be even more useful if you could override the feed.xml template completely? Then you could stick the feed.xml config at the root, point to that new path and completely customize the RSS feed the way you want if you want to.

lancecarlson:
Actually, after looking again.. this looks like this works already? If I have a feed.xml in my root of my project, it loads that instead?

parkr:
> If I have a feed.xml in my root of my project, it loads that instead?

I believe this is the case, yes. :+1:

pathawks:
Just spitballing here: What if we included the most recent 10 posts, _and_ all posts published in the past 48 hours? Would this solve the problem of feed readers missing posts because of infrequent polling?

benbalter:
@pathawks I like that idea. Not sure if it's possible in liquid or crazy complicated, but how about 10 posts OR everything in the past 24 hours, whichever is more?

pathawks:
@benbalter I will take a look tonight and see what is possible.

pathawks:
> What if we included the most recent 10 posts, and all posts published in the past 48 hours?

This seems to not be currently possible. jekyll/jekyll#4860

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least
one month.

The resources of the Jekyll team are limited, and so we are asking for your help.

[...]

PR #107, opened 2016-03-29
make the number of feed entries configurable

hackerkid:
This would solve https://github.com/jekyll/jekyll-feed/issues/56 by making the number of feed entries configurable. If the limit is not provided by the user the default value of 10 would be used.


pathawks:
This feature cannot be merged without a coresponding test.

I’m not sure if it was ever decided if the additional config option would be worth the reduced simplicity to the end user.

pathawks:
Oh, but thank you for the PR! :+1:

Didn’t mean to sound so negative there

hackerkid:
I have added the test. Looking for the Jekyll team's call on this feature. :sparkles:

benbalter:
@hackerkid thanks for the PR. See the discussion in https://github.com/jekyll/jekyll-feed/issues/56, which was never fully resolved. What's your use case for submitting this pull request? While technically feasible, if the number of posts is arbitrary, we might as well pick a good arbitrary number, and save the user the complexity.

hackerkid:
@benbalter I don't have a good use case. But I went through the source of some Jekyll powered blogs. The common feed limits were 10, 15 and 20.  There was no majority for any feed limit. We may be able to find a majority if we increase the number of blogs analysed. 

Jekyll Blog | Feed Limit
------------ | -------------
[feross.org](https://github.com/feross/feross.org) | 20
[mojombo.github.io](mojombo.github.io) | No Limit
[The Pug Automatic](http://thepugautomatic.com/) | 20
[vrepin.org](https://github.com/vitalyrepin/vrepinblog) | 12
[Paper Plane](http://www.paperplanes.de/) | No Limit
[Tagaholic Me](http://tagaholic.me/) | 15
[Tagaholic](favrik.com) | 15
[http://9leg.com/](zJiaJun.github.io) | 10


Wordpress has a built in feature to change the feed limit. Users migrating from Wordpress to Jekyll may like to have a control over the feed limit.

GarthDB:
@benbalter I could use this for the [Open Design Foundation blog](https://github.com/DesignOpen/designopen.github.io/). We have a lot of small resources that end up in the feed as well, which would push out the articles. We are currently using a custom [generated file](https://github.com/DesignOpen/designopen.github.io/blob/master/atom.xml) which gets the job done, but the plugin would be cleaner.

To save the user complexity just make a default value to fall back on if the option isn't in the config file. Right now it is 10 which seems like a good place to start.

GarthDB:
Looking at @hackerkid's commit that's exactly what he did. If this were merged it would add zero complexity to anyone already using the plugin and add value to anyone looking for this functionality.

benbalter:
> If this were merged it would add zero complexity to anyone already using the plugin 

@GarthDB I disagree. We'd be required to document it, which means it's another thing you have to read when you're getting started. It also pushes a decision, albeit an optional one, on the user. As I'm setting up my site, should the limit be 5? 10? 15? 20? How do I know? 

> We have a lot of small resources that end up in the feed as well

Thanks for sharing that use case. Very helpful. I have two questions:

1. Do you think the [*majority* of users](http://ben.balter.com/2016/03/08/optimizing-for-power-users-and-edge-cases/) would use the plugin similarly?

2. Have you found that items were not being picked up by feed readers and that you had to increase the limit? If so, what was that number, and if reasonable for all sites, can we just change the default? Is there any evidence on how often feed readers poll?

GarthDB:
@benbalter I'm a fan of the attitude of erring on the side of simplicity; it's easy for things to get out of hand.

Your first question kinda shows why it be good to have a different limit based on the project and content. If there isn't an obvious universal number to set as the limit it could be beneficial to let the users have the ability to decide. I don't see it as pushing the decision to them as much as giving them the ability to decide if they want to.

1. Not exactly the same way - but if they aren't looking for it I think the majority of users wouldn't mind the extra paragraph it would add to the readme.
2. For our exact use case it actually depends on what the reader is coming to find. Since new content is published infrequently and most of it is fairly timeless it makes sense to have a long list for new subscribers and it doesn't have a large impact on existing subscribers.

Again, I understand the benefits of simplicity, but Jekyll is a developer aimed tool. We have a marketing manager who writes posts for the [PhoneGap Blog](https://github.com/phonegap/blog), but she had to learn git just to publish posts. The barrier of entry is already huge for anyone not familiar with the basic tool chain. A plugin like this should help streamline and standardize common workflows. If the plugin doesn't have some flexiblity to adapt to the content of the individual project, developers will stick with building custom, less standard solutions.

It's up to you to decide what is valuable to GitHub and it's users, but if you are basing all your decisions on what the majority of users will want, you'll need to do constant indepth user research because you won't know if people would use something if it is never an option. Davatron5000 makes a [good point](https://github.com/github/pages-gem/pull/106#issuecomment-175271595) that I think is somewhat in the same vein.

hackerkid:
@benbalter I am closing this.

yyyc514:
In the hopes that it helps someone (it's easy to do this without a plugin at all):

http://runningblind.com/2018/12/22/jekyll-jekyll-feed-you-don-t-need-it/

That's the route I ended up going. And then tuning the number is just changing the number in your own template.

Issue #108, opened 2016-04-17
Issue with relative URLs


sondr3:
I just upgraded from 0.4.0 to 0.5.0 and suddenly I get this error: 

```
Liquid Exception: both URI are relative in _includes/head.html, included in _layouts/default.html
jekyll 3.1.2 | Error:  both URI are relative
```

I figured out that it's because you need to have `http://` in front of your URL before, however you didn't need to in 0.4.0. Just curious if this was a intended change.

pathawks:
> I figured out that it's because you need to have http:// in front of your URL before, however you didn't need to in 0.4.0. 

Before what?

How can I reproduce this error? What does your `_includes/head.html` contain?

sondr3:
Sorry, now you need to have your URL include `http://` in front of the URL in `_config.yml` whereas before 0.5.0 you didn't need. All you need to do to reproduce is add `jekyll-feed`, add the feed meta and then have your URL be `url: www.example.com` or `url: http://www.example.com`. Before the new version the first way to write your URL worked.

pathawks:
I have figured it out!

We are [now using 'URI.join' to join `config["url"]` and `config["baseurl"]`](https://github.com/jekyll/jekyll-feed/compare/v0.4.0...v0.5.0#diff-2ca45c54016b64d06027be90e8c00fd6R34) which won't work if neither contains a scheme.

This _shouldn't_ have worked to begin with. If you don't at least include a `//` in the URL, it will not resolve properly.

I agree with you that the error message (`Liquid Exception: both URI are relative`) does not contain enough information to understand how to fix the problem. Maybe a `try catch` with a more helpful warning would make things easier?

sondr3:
Thanks, good to know :)

PR #109, opened 2016-04-17
Rethrow error on invalid site.url

pathawks:
#108

Definitely need a more descriptive error message. Am I on the right track with this?

### DIFF
--- a/lib/jekyll/errors/jekyll-feed.rb
+++ b/lib/jekyll/errors/jekyll-feed.rb
@@ -1,0 +1,7 @@
+module Jekyll
+  module Errors
+    module JekyllFeed
+      InvalidURLConfigurationError = Class.new(Jekyll::Errors::FatalException)
+    end
+  end
+end
--- a/lib/jekyll/feed_meta_tag.rb
+++ b/lib/jekyll/feed_meta_tag.rb
@@ -31,10 +31,25 @@ module Jekyll
 
     def url
       if config["url"]
-        URI.join(config["url"], config["baseurl"])
+        begin
+          URI.join(config["url"], config["baseurl"])
+        rescue
+          raise Jekyll::Errors::JekyllFeed::InvalidURLConfigurationError, url_error
+        end
       elsif config["github"] && config["github"]["url"]
         config["github"]["url"]
       end
+    end
+
+    def url_error
+      <<-EOS
+One or both of `url` and `baseurl` are invalid in your configuration file. We were unable to create a valid URL with them:
+
+    url = #{config["url"].inspect}
+    baseurl = #{config["baseurl"].inspect}
+
+Please correct them, or remove them from your configuration.
+      EOS
     end
 
     def title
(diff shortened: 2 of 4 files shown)


lib/jekyll/feed_meta_tag.rb
```diff
@@ -31,7 +31,13 @@ def path
 
     def url
       if config["url"]
-        URI.join(config["url"], config["baseurl"])
+        begin
+          URI.join(config["url"], config["baseurl"])
+        rescue
+          raise Jekyll::Errors::FatalException, <<-MSG
+`site.url` must be an absolute URL
```
pathawks: This should be something a whole lot more descriptive 😒

parkr:
@pathawks Great! I'd create your own custom error for this and output the values you have in your `URI.join` arguments in the error message.

```ruby
Jekyll::FeedMetaTag::InvalidURLConfigurationError = Class.new(Jekyll::Errors::FatalException)
```

or something. Then 

```ruby
raise Jekyll::FeedMetaTag::InvalidURLConfigurationError, <<-EOS
One or both of `url` and `baseurl` are invalid in your configuration file. We were unable to create a valid URL with them:

    url = "#{url.inspect}"
    baseurl = "#{baseurl.inspect}"

Please correct them, or remove them from your configuration.
EOS
```

pathawks:
How does this look? 👀

lib/jekyll/feed_meta_tag.rb
```diff
@@ -31,12 +31,27 @@ def path
 
     def url
       if config["url"]
-        URI.join(config["url"], config["baseurl"])
+        begin
+          URI.join(config["url"], config["baseurl"])
+        rescue
+          raise Jekyll::Errors::JekyllFeed::InvalidURLConfigurationError, url_error
```
parkr: Why this namespacing?

lib/jekyll/errors/jekyll-feed.rb
```diff
@@ -0,0 +1,7 @@
+module Jekyll
+  module Errors
+    module JekyllFeed
```
parkr: Why this namespacing?

spec/jekyll-feed_spec.rb
```diff
@@ -224,6 +224,14 @@
         expect(feed_meta).not_to include('title=')
       end
     end
+
+    context "with an invalid site URL" do
+      let(:overrides) { { "url" => "example.org" } }
```
parkr: Why does this break it?

lib/jekyll/errors/jekyll-feed.rb
pathawks: Because I have no idea what I am doing.
You think `JekyllFeed::Errors` ?

spec/jekyll-feed_spec.rb
pathawks: Because there is no `http://` or `https://`

If this appeared in an `href`, it would be interpreted as `./example.org` which breaks things.

pathawks:
If we could just use Jekyll's new URL filters, we'd all be a lot better off.

pathawks:
Now that we are using Jekyll's url filters, I think this is a Jekyll issue.

PR #110, opened 2016-04-18
Fix mangling of whitespace when `site.lang` is set

benbalter:
Which slams `hreflang` into the preceding attribute, causing parse errors.

This just adds a quick negative look behind to change the whitespace stripping behavior.

### DIFF
--- a/lib/jekyll/jekyll-feed.rb
+++ b/lib/jekyll/jekyll-feed.rb
@@ -32,7 +32,8 @@ module Jekyll
 
     def feed_content
       feed = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", path)
-      feed.content = File.read(source_path).gsub(/\s+([{<])/, '\1')
+      content = File.read(source_path).gsub(/(?<!\")\s+([<{])/, '\1')
+      feed.content = content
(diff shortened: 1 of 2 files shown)


lib/jekyll/jekyll-feed.rb
```diff
@@ -32,7 +32,7 @@ def source_path
 
     def feed_content
       feed = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", path)
-      feed.content = File.read(source_path).gsub(/\s+([{<])/, '\1')
+      feed.content = File.read(source_path).gsub(/(?<!\")\s+([<{])/, '\1')
```
parkr: Would it have also worked to put the space on the other side of the `hreflang`, i.e. at the beginning (right after `{% if %}`) instead of at the end?

benbalter: There's no way we can remember to always do that going forward, nor could we expect contributors to deal with an odd behavior like that.

parkr: It's not ideal, true. 👍

parkr:
YOU'RE A HERO thank you

@jekyllbot: merge

PR #111, opened 2016-04-26
Added test for _data/authors.yml data and corrected mistakes

GarthDB:
While trying to use referenced author data in the `_data/authors.yml` file as outlined in the documentation I found a couple mistakes in the [`lib/feed.xml`](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml) file:

* The `site.data.authors[author].uri` is being output in `<email>` [#L70](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml#L70)
* The `site.data.authors[author].name` is not being set, so the reference itself is output in `<name>` [#L74](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml#L74)

I also added a corresponding test with a `_data/authors.yml` file and post, with an author reference, to the `spec/fixtures` directory.

PR #113, opened 2016-04-26
Cleanup `post_author` logic

pathawks:
As pointed out by @GarthDB in #111, the way we were doing it previously was a mess that camouflaged bugs. This builds on PR, further simplifying things and hopefully giving bugs fewer places to hide.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -67,11 +67,15 @@
           {% assign post_author_uri = post_author.uri %}
         {% else %}
           {% if site.data.authors and site.data.authors[post_author] %}
-            {% assign post_author_email = site.data.authors[post_author].uri %}
+            {% assign post_author_uri = site.data.authors[post_author].uri %}
           {% endif %}
         {% endif %}
         {% if post_author.name %}
           {% assign post_author = post_author.name %}
+        {% else %}
+          {% if site.data.authors and site.data.authors[post_author] %}
+            {% assign post_author = site.data.authors[post_author].name %}
+          {% endif %}
         {% endif %}
         <author>
             <name>{{ post_author | xml_escape }}</name>
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -53,30 +53,12 @@
       <content type="html" xml:base="{{ post.url | prepend: url_base | xml_escape }}">{{ post.content | strip | xml_escape }}</content>
 
       {% assign post_author = post.author | default: post.authors[0] | default: site.author %}
-      {% assign post_author_email = false %}
-      {% assign post_author_uri = false %}
+      {% assign post_author = site.data.authors[post_author] | default: post_author %}
+      {% assign post_author_email = post_author.email | default: nil %}
+      {% assign post_author_uri = post_author.uri | default: nil %}
+      {% assign post_author = post_author.name | default: post_author %}
+
       {% if post_author %}
-        {% if post_author.email %}
-          {% assign post_author_email = post_author.email %}
-        {% else %}
-          {% if site.data.authors and site.data.authors[post_author] %}
-            {% assign post_author_email = site.data.authors[post_author].email %}
-          {% endif %}
-        {% endif %}
-        {% if post_author.uri %}
-          {% assign post_author_uri = post_author.uri %}
-        {% else %}
-          {% if site.data.authors and site.data.authors[post_author] %}
-            {% assign post_author_uri = site.data.authors[post_author].uri %}
-          {% endif %}
-        {% endif %}
-        {% if post_author.name %}
-          {% assign post_author = post_author.name %}
-        {% else %}
-          {% if site.data.authors and site.data.authors[post_author] %}
-            {% assign post_author = site.data.authors[post_author].name %}
-          {% endif %}
-        {% endif %}
         <author>
             <name>{{ post_author | xml_escape }}</name>
           {% if post_author_email %}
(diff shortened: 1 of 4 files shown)


GarthDB:
Much cleaner. :+1:

pathawks:
Could I get an extra set of eyes from @parkr or @benbalter please?

lib/feed.xml
```diff
@@ -53,26 +53,12 @@
       <content type="html" xml:base="{{ post.url | prepend: url_base | xml_escape }}">{{ post.content | strip | xml_escape }}</content>
 
       {% assign post_author = post.author | default: post.authors[0] | default: site.author %}
-      {% assign post_author_email = false %}
-      {% assign post_author_uri = false %}
+      {% assign post_author = site.data.authors[post_author] | default: post_author %}
```
parkr: Should this be `post_author_data` not to be confused with `post_author` which is just the name?

parkr:
Excellent work! 👍

lib/feed.xml
pathawks: `post.author` can be a string or a hash. It's kind of a mess ¯\_(ツ)_/¯

If we changed _this_ line, we would also need to change the line above it to `post_author_data`, but then it would seem weird to `{% assign post_author_data = site.data.authors[post_author_data] | default: post_author_data %}`

I am open to suggestion though.

GarthDB: Maybe if it is just a string, and not a reference, we assign it to author name.

lib/feed.xml
```diff
@@ -53,26 +53,12 @@
...
+      {% assign post_author_email = post_author.email | default: false %}
```
benbalter: What's the rationale for having a `false` default, rather than `nil`?

lib/feed.xml
benbalter: Wondering if, between this and Jekyll SEO Tag, it may make sense to have a shared `jekyll-author` gem, that would never be included directly (I *guess* it could?), but would handle author being a string, a hash, or a ref to `_data`, and would always return an object, maybe in Rubyland?

GarthDB: @benbalter that's a great idea, but it would probably be best to start by releasing a patch that fixes the documented functionality first.

lib/feed.xml
pathawks: You are right, `nil` makes more sense.

lib/feed.xml
pathawks: > Maybe if it is just a string, and not a reference, we assign it to author name.

I don't know of a way in Liquid to see what type a variable is.

> it may make sense to have a shared `jekyll-author` gem

Totally agree, but as @GarthDB mentioned, let's get this bug fix out first.

GarthDB: I'm on the road so I can't write out a full example, but just checking if the string is a key of the authors object and has a value, otherwise assume the string is a name.

benbalter: > but it would probably be best to start by releasing a patch that fixes the documented functionality first.

Completely agree.

lib/feed.xml
```diff
@@ -53,28 +53,14 @@
       <content type="html" xml:base="{{ post.url | prepend: url_base | xml_escape }}">{{ post.content | strip | xml_escape }}</content>
 
       {% assign post_author = post.author | default: post.authors[0] | default: site.author %}
-      {% assign post_author_email = false %}
-      {% assign post_author_uri = false %}
+      {% assign post_author = site.data.authors[post_author] | default: post_author %}
+      {% assign post_author_email = post_author.email | default: nil %}
+      {% assign post_author_uri = post_author.uri | default: nil %}
+      {% assign post_author_name = post_author.name | default: post_author %}
```
pathawks: Does this make more sense? (See also line [63](#diff-93e2ef419e8edc5867dde4edec770ec8R63))

GarthDB:
@parkr @pathawks @benbalter while waiting for this I decided to work on a [`jekyll-authors`](https://github.com/GarthDB/jekyll-authors) plugin

It should provide what would be needed by `jekyll-feed` and [`jekyll-seo-tag`](https://github.com/jekyll/jekyll-seo-tag) in the form of Jekyll hooks.

Is this something you'd be interested in using? If so, I'm happy to transfer the repo if needed and make pull requests to the 2 plugins.

I did find out that the gem name `jekyll-authors` is already taken by a gem that generates author pages. Not sure how we would proceed to name and publish it.

pathawks:
Is _this_ PR good to go, or is there still something left to modify?

lib/feed.xml
benbalter: How does the behavior change if you remove the ` | default: nil`?

benbalter:
> Is this PR good to go, or is there still something left to modify?

One style question, otherwise, :+1: (and agree we should merge this before looking into a more robust solution).

>  while waiting for this I decided to work on a jekyll-authors plugin

That's awesome. Thanks @GarthDB. One thing to note, that moves things from Liquid to Ruby, which isn't necessarily a bad thing, but makes things a bit trickier in terms of security.

GarthDB:
@benbalter is there a more secure way to do it that both jekyll-seo-tag and jekyll-feed could use?

I figured jekyll hooks would be the best way to prep the data before either could use it.

lib/feed.xml
pathawks: The behavior is the same either way, but I feel like including some sort of `default` is an acknowledgment that this may not be set, and that is ok.

If you do not think it makes the code easier to follow, I can remove it.

benbalter: > The behavior is the same either way, but I feel like including some sort of default is an acknowledgment that this may not be set, and that is ok.

👍

pathawks:
@jekyllbot: merge

GarthDB:
Thanks for the work @pathawks

PR #117, opened 2016-05-20
Use smartify filter

pathawks:
Requires Jekyll 3.1

Processes most strings with SmartyPants to replace characters with their fancy typographic equivalent.

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,11 +112,15 @@ There are several ways to convey author-specific information. Author information
   author: benbalter
   ```
 
-* `image` - URL of an image that is representative of the post.  
+* `image` - URL of an image that is representative of the post.
 
 ### Meta tags
 
 The plugin exposes a helper tag to expose the appropriate meta tags to support automated discovery of your feed. Simply place `{% feed_meta %}` someplace in your template's `<head>` section, to output the necessary metadata.
+
+### SmartyPants
+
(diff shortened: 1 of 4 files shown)


benbalter:
What's the value of using Smartypants here? Aren't we actually XML escaping, since we're generating XML?

pathawks:
If there is a title like `Pat's Blog`, the `smartify` filter will substitute a real apostrophe and escape, so the output is `Pat&lsquo;s Blog`

benbalter:
Nice. 👍. Also double checked, per https://validator.w3.org/feed/docs/atom.html, that HTML entities are allowed for plain text fields.

pathawks:
I was wrong, `smartify` does not escape it's output. So with input `Pat's Blog`, it will output `Pat’s Blog`.

With this in mind, we probably want to keep `xml_escape` in there?

README.md
```diff
@@ -112,12 +112,16 @@ There are several ways to convey author-specific information. Author information
   author: benbalter
   ```
 
-* `image` - URL of an image that is representative of the post.  
+* `image` - URL of an image that is representative of the post.
```
pathawks: Sorry for the noise. My text editor strips trailing space on save 😕

pathawks:
@jekyllbot: merge

Issue #118, opened 2016-06-10
Stylesheets with XSLT


jhabdas:
XML supports stylesheets for styling documents using XSLT ([example](http://www.ibm.com/developerworks/xml/tutorials/x-atomxsl/x-atomxsl.html)). Web feeds are usually boring. Let's make them fun again by creating a config option to set a stylesheet `href`!

```xml
<?xml version="1.0" encoding="utf-8"?>
<?xml-stylesheet type="text/xml" href="tickertape.xslt.xml"?>
```

benbalter:
@jhabdas that's for opening the idea. I have two questions:

1. How often to users view the feed XML in their browser? What's advantage does a stylesheet provide?

2. What's the advantage to making the XSLT configurable, versus including a default XSLT?

jhabdas:
1) Rarely. But themes like Casper make feeds prominent (upper right of a page) and, as a result, users are more likely to view an unstyled XML feed.

2) The idea of making the XSLT configurable is to allow some to apply creative expression to their web feeds to delight users much the same way 404 pages often get special treatment. A default could be useful as well, but keeping it off means less issues and maintenance for the plug-in.

benbalter:
> A default could be useful as well, but keeping it off means less issues and maintenance for the plug-in.

What if, similar to the feed itself, we wrote a `feed.xslt` file to the site destination, unless it already exists, providing a styled default, but also allowing any user to add their own `feed.xslt` file, without requiring a configuration flag?

jhabdas:
SGTM

benbalter:
@jhabdas if you're able to, any interest in starting a pull request for this? (Even if just to add a default XSLT?)

jhabdas:
On it

jhabdas:
Cloned and tried to run the default rake task after bundling but hit friction building on the version of Ruby Travis is using to run your CI builds. Rather than fighting it (I'm on OS X) or setting up a Docker environment (which would work if one was available) I decided instead to run a quick test using the IBM Ticker Tape XSLT example (linked earlier) against @pathawks Atom feed with the following result:

![screen shot 2016-06-10 at 1 57 33 pm](https://cloud.githubusercontent.com/assets/440298/15975474/4692ee9a-2f13-11e6-8a1d-3914d6e4f927.png)

**Source files used:**

[tickerhawk.zip](https://github.com/jekyll/jekyll-feed/files/309597/tickerhawk.zip)

**Testing:**

[...]

PR #119, opened 2016-06-10
Add XML stylesheet example with XSLT

jhabdas:
Just an example for #118.

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<?xml-stylesheet type="text/xml" href="{{ url_base }}/feed.xslt.xml"?>
 {% if site.url %}
   {% assign url_base = site.url | append: site.baseurl %}
 {% else %}
--- a/lib/jekyll/jekyll-feed.rb
+++ b/lib/jekyll/jekyll-feed.rb
@@ -12,6 +12,9 @@ module Jekyll
       unless feed_exists?
         @site.pages << feed_content
       end
+      unless xslt_exists?
+        @site.pages << xslt_content
+      end
     end
 
     private
@@ -45,6 +48,31 @@ module Jekyll
         File.exist? @site.in_source_dir(path)
       else
         File.exist? Jekyll.sanitized_path(@site.source, path)
+      end
+    end
+
+    # Path to the feed.xslt.xml template file
+    def xslt_source_path
+      File.expand_path "../feed.xslt.xml", File.dirname(__FILE__)
+    end
+
+    def xslt_content
+      xslt_path = "feed.xslt.xml"
+      xslt = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", xslt_path)
+      xslt.content = File.read(xslt_source_path).gsub(/(?<!\")\s+([<{])/, '\1')
+      xslt.data["layout"] = nil
+      xslt.data["sitemap"] = false
+      xslt.output
+      xslt
+    end
+
+    # Checks if a feed stylesheet exists in the site source
+    def xslt_exists?
+      xslt_path = "feed.xslt.xml"
+      if @site.respond_to?(:in_source_dir)
+        File.exist? @site.in_source_dir(xslt_path)
+      else
+        File.exist? Jekyll.sanitized_path(@site.source, xslt_path)
       end
     end
   end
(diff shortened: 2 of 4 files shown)


lib/feed.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<?xml-stylesheet type="text/xml" href="feed.xslt.xml"?>
```
benbalter: This should be `{{ url_base}}/feed.xslt.xml`

lib/feed.xml
```diff
@@ -1,4 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
+<?xml-stylesheet type="text/xml" href="{{ url_base }}/feed.xslt.xml"?>
```
jhabdas: Updated to use `url_base` liquid object.

jhabdas:
Working on some basic customizations to make the default XSLT template look a little nicer and be more usable.

jhabdas:
Made a small adjustment to the XSLT stylesheet so it outputs both the published and updated dates, and fixed a unit test for `url_base` which I may have missed an edge case for.

Working with the XSLT is tricky so I opted not to go too much further. Ideally some basic date parsing via use of [`format-dateTime`](https://www.w3.org/TR/xpath-functions-30/#func-format-dateTime) could be helpful for displaying dates. If someone wanted to go all out they could adapt [the code](https://plugins.svn.wordpress.org/xslt/trunk/) for [Better RSS Feeds](https://wordpress.org/plugins/xslt/screenshots/) to output a fully styled HTML document with hyperlinks and social sharing. For my purposes I'm happy with simply not seeing unformatted XML.

benbalter:
> Made a small adjustment to the XSLT stylesheet so it outputs both the published and updated dates, and fixed a unit test for url_base which I may have missed an edge case for.

Would you mind posting a screenshot?

Also /cc @jglovier famed Jekyll designer, in case he has any suggestions.

lib/jekyll/jekyll-feed.rb
```diff
@@ -47,5 +50,30 @@ def feed_exists?
...
+      xslt = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", xslt_path)
+      xslt.content = File.read(xslt_source_path).gsub(/(?<!\")\s+([<{])/, '\1')
+      xslt.data["layout"] = nil
+      xslt.data["sitemap"] = false
+      xslt.output
+      xslt
+    end
+
+    # Checks if a feed stylesheet exists in the site source
+    def xslt_exists?
```
benbalter: Should this an the `feed_exists?` methods be abstracted into a `file_exists?(file)` method could be shared for both?

lib/jekyll/jekyll-feed.rb
```diff
@@ -47,5 +50,30 @@ def feed_exists?
         File.exist? Jekyll.sanitized_path(@site.source, path)
       end
     end
+
+    # Path to the feed.xslt.xml template file
+    def xslt_source_path
+      File.expand_path "../feed.xslt.xml", File.dirname(__FILE__)
+    end
+
+    def xslt_content
```
benbalter: Same with `content_for_file(file)` or similar. Just trying to DRY things up where we can.

spec/jekyll-feed_spec.rb
```diff
@@ -249,6 +249,12 @@
     end
   end
 
+  context "feed stylesheet" do
```
benbalter: We'll also want to test that the resulting site includes the stylesheet file.

jhabdas:
Here's a screenshot for the basic example provided. It made it immediately clear I wasn't outputting my RSS updated/modified dates properly from my Jekyll site:

screen shot 2016-06-13 at 2 46 04 pm.

I'll take a stab at those refactorings.

spec/jekyll-feed_spec.rb
jhabdas: Could you provide a little more direction on this one? I believe I was able to tackle the other two refactorings.

jhabdas: Disregard this request. I think I got it.

jhabdas:
@benbalter Ready for another review pass.

benbalter:
@pathawks @parkr any strong feelings here?

pathawks:
LGTM :shipit:

parkr:
Fine with me!

benbalter:
@jekyllbot: merge

benbalter:
@jhabdas Thanks for this!

jhabdas:
My pleasure. This was my second time writing teh Ruby. Thanks for making it fun @benbalter. This is almost better than humans.txt!

PR #127, opened 2016-08-30
Use type="html" to skirt around double escaping problem

### LINKED ISSUE #98: Ampersands in a blog summary cause invalid atom
Hey,

If I put an ampersand in the leading paragraph of a blog post, it becomes part of the `<summary>` tag which causes the [W3 validator](http://validator.w3.org/feed/check.cgi) to fail with this message:

```
summary should not contain HTML unless declared in the type attribute:
&amp;
```

Here's a commit which demonstrates "fixing" a blog post by removing the ampersands:

https://github.com/paulfurley/paulfurley.github.io/commit/7bbc20754353b0277fdec7c6193516195fa06f22

Cheers!

pathawks:
I really wanted to find a way to just avoid double-escaping the entities, but I have not. It seems this will work instead.

Fixes #98

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -13,9 +13,9 @@
   <id>{{ url_base | xml_escape }}/</id>
 
   {% if site.title %}
-    <title>{{ site.title | smartify | xml_escape }}</title>
+    <title type="html">{{ site.title | smartify | xml_escape }}</title>
   {% elsif site.name %}
-    <title>{{ site.name | smartify | xml_escape }}</title>
+    <title type="html">{{ site.name | smartify | xml_escape }}</title>
   {% endif %}
 
   {% if site.description %}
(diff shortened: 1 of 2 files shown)


xHN35RQ:
I just realized the [feed title](https://github.com/jekyll/jekyll-feed/blob/567b1f8fbda791f4ec15270db17f9d30081e1dc3/lib/feed.xml#L15-#L19) should also have `type="html"`. Right now it is defaulting to `type="text"` and thus the double encoded entities are appearing in the site title as well. Using the following config:

```
title: Site title & < >
description: Blog description & < >
```

produces the following effect when previewing the feed in Firefox:

screen shot 2016-08-30 at 01 48 20

parkr:
This LGTM. ❤️ I agree it should also be in the title, @xHN35RQ.

lib/feed.xml
```diff
@@ -13,9 +13,9 @@
   <id>{{ url_base | xml_escape }}/</id>
 
   {% if site.title %}
-    <title>{{ site.title | smartify | xml_escape }}</title>
+    <title type="html">{{ site.title | smartify | xml_escape }}</title>
   {% elsif site.name %}
```
pathawks: Note to self: refactor to make use of `default` filter.

pathawks:
I have added `type="html"` to the feed title :+1:

pathawks:
...and I have updated the test to expect `type="html`. I'm actually a bit surprised that we only broke one test.

At any rate, should be good now.

parkr:
LGTM!

benbalter:
LGTM.

pathawks:
@jekyllbot merge 🎉

pathawks:
@jekyllbot: merge 🎉

PR #130, opened 2016-09-21
add support for categories frontmatter

### LINKED ISSUE #129: support both category and categories front-matter
"out of the box" Jekyll supports both `category` and `categories`: https://jekyllrb.com/docs/frontmatter/#predefined-variables-for-posts

Could update the line [here](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml#L74) to support `categories` front matter.

digitalvapor:
resolves #129

### DIFF
--- a/lib/feed.xml
+++ b/lib/feed.xml
@@ -74,6 +74,9 @@
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
       {% endif %}
+      {% for category in post.categories %}
+        <category term="{{ category | xml_escape }}" />
+      {% endfor %}
 
       {% for tag in post.tags %}
         <category term="{{ tag | xml_escape }}" />


pathawks:
> `atom:entry` elements MAY contain any number of `atom:category` elements.

Rather than include one or the other, I guess we should include both `post.category` and _all_ `post.categories`.

Oh Jekyll, sometimes I must try so hard to love you 😏

lib/feed.xml
```diff
@@ -73,6 +73,8 @@
 
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
+      {% elsif post.categories %}
+        <category term="{{ post.categories | first | xml_escape }}" />
```
pathawks: We should include _all_ categories, not just the first.

digitalvapor:
Sure! I'll update.

digitalvapor:
updated. included fix for #131 too.

lib/feed.xml
```diff
@@ -73,8 +73,11 @@
 
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
-      {% elsif post.categories %}
-        <category term="{{ post.categories | first | xml_escape }}" />
+      {% endif %}
+      {% if post.categories %}
+        {% for category in post.categories %}
+        <category term="{{ category | first | xml_escape }}" />
```
pathawks: I don't think we need `first`, since we will only have one category.

lib/feed.xml
```diff
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="utf-8"?>
-<?xml-stylesheet type="text/xml" href="{{ url_base }}/feed.xslt.xml"?>
```
pathawks: Whoops! Thanks for noticing.

For the benefit of the changelog, could you punt this in a separate PR please?

lib/feed.xml
```diff
@@ -74,6 +74,11 @@
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
       {% endif %}
+      {% if post.categories %}
+        {% for category in post.categories %}
+        <category term="{{ category | first | xml_escape }}" />
```
pathawks: I don't think we need `first`, as we're only working with one category at a time.

lib/feed.xml
```diff
@@ -74,6 +74,11 @@
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
       {% endif %}
+      {% if post.categories %}
```
pathawks: We shouldn't need to wrap the `for` loop with an `if`. If `post.categories` is empty, the `for` loop will execute zero times :+1:

lib/feed.xml
digitalvapor: fixed

lib/feed.xml
digitalvapor: removed

digitalvapor:
fixed and squashed commits

pathawks:
I would love to get one more LGTM from @jekyll/plugin-core

digitalvapor:
to be fullproof I suppose `category` should be pushed into the `categories` array and ensured that the array is unique and just iterate over the categories.

parkr:
@digitalvapor I think [Jekyll does that already](https://github.com/jekyll/jekyll/blob/06651c960aff4173121fa934b741b764bbe1a4e8/lib/jekyll/document.rb#L467-L471) which is why I bring up the concern.

pathawks:
> Jekyll does that already which is why I bring up the concern.

That means we could just get rid of lines 74-76 and leave this new `for` loop, which is awesome.

Trouble is, we don't currently have _any_ tests for `<category>`.

@digitalvapor Would you mind adding to the `fixture` site a post with a `category` and a post with `categories` set in the front matter and adding a test to make sure that both output correctly?

digitalvapor:
sure!

pathawks:
I was hoping to clean this up to get it ready to merge, but the branch has been deleted❗️😑

Will instead get this across the finish line in #153

nebulousdog:
thanks for following up @pathawks. I think it was because I changed my username :man_facepalming:.

PR #134, opened 2016-09-24
Use filters to clean up Liquid template

pathawks:
Obviously, we won't be able to ship this until Jekyll 3.3 lands, but I was so excited about the new filters I wanted to see how we could use them to clean up our template.

We will need to drop support for anything earlier than Jekyll 3.3, but I feel the advantages are worth it. Perhaps we wait until GitHub Pages supports Jekyll 3.3?

jekyll/jekyll#5399


parkr:
> Perhaps we wait until GitHub Pages supports Jekyll 3.3?

:+1:

benbalter:
@pathawks can you merge? If so, I'm fine if JekyllBot merges this, if you'd like to follow up with another PR to fix those rubocop offenses (welcome to tackle them here as well if you'd prefer).

pathawks:
@jekyllbot: merge

PR #136, opened 2016-09-28
Do not override `site.time`

### LINKED ISSUE #135: Draft articles disappear on regeneration with jekyll-feed enabled
Steps to reproduce:
1. Start a new Jekyll site with 3.2.1
2. Install and enable jekyll-feed
3. Create a draft post in _drafts
4. Serve the site with: `jekyll s --drafts`
5. Notice that your draft is listed on the front page.
6. Edit your draft. The site should regenerate.
7. Your draft is no longer listed.

This does not happen with non drafts. Additionally, if you edit some other file (like a post) the draft will reappear on site regeneration. Only the most recently edited draft disappears.

pathawks:
Fixes #135

### DIFF
--- a/lib/jekyll/jekyll-feed.rb
+++ b/lib/jekyll/jekyll-feed.rb
@@ -8,7 +8,7 @@ module Jekyll
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
-      @site.config["time"] = Time.new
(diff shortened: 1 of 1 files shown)


pathawks:
This may not be so simple. We need to avoid setting `site.time`, not just `site.config['time']`

PR #141, opened 2016-10-15
Simplify minify regular expression

pathawks:


benbalter:
I see a `| minify` filter in our future.

pathawks:
> I see a `| minify` filter in our future.

Not a bad idea, but the trick is that is kind of has to be a context aware minify. It is easy to minify these templates before they are rendered because we _know_ that there is nowhere in the template where whitespace is significant. Once we start dealing with user generated content, we never know when we might run into a `<pre>` or similar where we cannot just blindly strip whitespace.

pathawks:
@jekyllbot: merge +dev

lib/jekyll/jekyll-feed.rb
```diff
@@ -18,6 +18,12 @@ def generate(site)
 
     private
 
+    # Matches all whitespace that follows
+    #   1. A '>', which closes an XML tag or
+    #   2. A '}', which closes a Liquid tag
+    # We will strip all of this whitespace to minify the template
+    MINIFY_REGEX = %r!(?<=>|})\s+!
```
pathawks: This _could_ be changed to `%r!(?<=[>}])\s+!`

I don't know if that would be any better or worse 🐳

PR #143, opened 2016-10-20
Use new `normalize_whitespace` filter

pathawks:
[`normalize_whitespace`](https://github.com/jekyll/jekyll/blob/v3.2.0/lib/jekyll/filters.rb#L153-L160) will replace any run of whitespace (including newlines) with a single space, and strip all space from the beginning and end. :+1:


pathawks:
@jekyllbot: merge +minor

PR #150, opened 2016-12-30
Use absolute_url to generate the feed_meta url

benbalter:
Rather than writing our own URL logic, lets just let Jekyll do the heavy lifting for us.

This should be a backwards compatible change.


benbalter:
@jekyllbot: merge +minor.

PR #151, opened 2016-12-30
Namespace as JekyllFeed

benbalter:
This PR moves the plugin from the Jekyll namespace to its own JekyllFeed namespace to match the design pattern of the other Jekyll plugins we maintain.

There should be no substantive changes here... just renaming e.g, `Jekyll::FeedMetaTag` to `JekyllFeed::MetaTag`, to (A) provide parity with other plugins and (B) avoid potential namespace conflicts.


benbalter:
@jekyllbot: merge +dev.

PR #152, opened 2017-01-05
Feed entries must contain <author>

### LINKED ISSUE #146: Feeds must contain <author> to validate
as reported in https://github.com/jekyll/minima/issues/89:
* the generated feed is broken. https://validator.w3.org/feed reports a missing author element and other problems.
* also, it seems that feed.xml gets served as `application/xhtml+xml` but it should be `application/atom+xml`

pathawks:
If no author is set, this will output `<author><name></name></author>`.

Feed validator will _recommend_ not having a blank `name`, but this will validate where previously it would not, because `<author>` is a required child of `<entry>`

Fixes #146

/cc: @dret @zixia


pathawks:
@jekyllbot: merge +minor

PR #153, opened 2017-01-05
add support for categories frontmatter

### LINKED ISSUE #129: support both category and categories front-matter

pathawks:
Replaces #130

Fixes #129


pathawks:
I had to create a second fixture site because the feed is limited to 10 items, and the feed already contained 10 items 😂

spec/jekyll-feed_spec.rb
```diff
@@ -261,6 +261,33 @@
...
+      post = feed.items.last
+      expect(post.categories[0].term).to eql('single')
+      expect(post.categories.count).to eql(1)
+    end
+
+    it "outputs categories for posts with one `category`" do
+      post = feed.items.first
+      expect(post.categories[0].term).to eql('1')
+      expect(post.categories[1].term).to eql('2')
+      expect(post.categories[2].term).to eql('3')
```
pathawks: Perhaps I should name the categories `first`, `second`, `third`?

nebulousdog:
thanks @pathawks!

danilowoz:
Please, someone, merge it 🙏

spec/jekyll-feed_spec.rb
```diff
@@ -289,6 +289,33 @@
...
+      expect(feed.items.count).to eql(2)
+    end
+
+    it "outputs categories for posts with one `category`" do
+      post = feed.items.last
+      expect(post.categories[0].term).to eql('single')
+      expect(post.categories.count).to eql(1)
+    end
+
+    it "outputs categories for posts with one `category`" do
```
DirtyF: @pathawks you meant multiples categories?

nebulousdog:
thanks for pushing this forward all

n1k0:
Do we have any news with this? Would be nice being able to support the `categories` front matter property; right now configuring `feed: categories:` in `_config.yml` will generate empty feeds for post categories using `categories` instead of `category`, which is super misleading to newcomers.

SGOrava:
> Do we have any news with this? Would be nice being able to support the `categories` front matter property; right now configuring `feed: categories:` in `_config.yml` will generate empty feeds for post categories using `categories` instead of `category`, which is super misleading to newcomers.

Oh man, can you put this into README ?
This killed this plugin for me, unusable.

ashmaroli:
@pathawks I pushed changes to how this was tested. Do you have any issues with the current diff..?

jeffgable:
As @SGOrava mentioned, until this is merged and released, can you at least put it in the README that "categories" frontmatter is not supported?  Not knowing that just cost me two hours.

DirtyF:
@pathawks What we should do here? 
We query only the posts from the `page.category`, then we loop on `post.categories`? what about the `page.category`?

DirtyF:
@jekyllbot: merge +minor

nebulousdog:
wonderful

PR #154, opened 2017-01-05
Filter out drafts before limit

pathawks:
It seems like drafts (which are not included in the feed) should not count toward the 10 post limit.

In a worst case scenario, having 10 or more drafts could result in an empty feed, no matter how many published posts the site contains.


pathawks:
@jekyllbot: merge +bug

PR #155, opened 2017-01-15
Site author should utilize author reference

### LINKED ISSUE #112: Site author does not utilize author reference
Going through [`lib/feed.xml`](https://github.com/jekyll/jekyll-feed/blob/master/lib/feed.xml#L24) I noticed it generates the feed's main author tag only from author data set in the main config. Would there be any benefit to adding support for using data in a `_data/authors.yml` file to reduce the duplication of data that could potentially be out of sync?

If it would be worth it I'd be happy to make a pull request.

pathawks:
Fixes #112

/cc: @GarthDB

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -19,14 +19,19 @@
     <subtitle>{{ site.description | xml_escape }}</subtitle>
   {% endif %}
 
-  {% if site.author %}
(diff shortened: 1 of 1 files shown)


GarthDB:
Sweet - looking forward to getting this fixed.

parkr:
Tests should cover the following cases:

1. Where `site.authors` is `nil`
2. Where `site.authors` is an array
3. Where `site.authors` is a hash

Issue #158, opened 2017-02-16
Double // in feed.xml for site <id>


mmistakes:
## Steps to reproduce

1. `jekyll new test-site`
2. Add `url: http://test-site.com` to _config.yml
3. Build site.
4. Examine `_site/feed.xml` 

```xml
<?xml version="1.0" encoding="utf-8"?><?xml-stylesheet type="text/xml" href="/feed.xslt.xml"?>
  <feed xmlns="http://www.w3.org/2005/Atom">
    <generator uri="http://jekyllrb.com" version="3.4.0">Jekyll</generator>
    <link href="http://test-site.com/" rel="alternate" type="text/html" />
    <updated>2017-02-16T12:27:08-05:00</updated>
    <id>http://test-site.com//</id>
    <title type="html">Your awesome title</title>
    ...
```

This line in [`feed.xml`](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L10) seems to be the culprit and outputs `//` for site root. Perhaps the combination of the new `absolute_url` `xml_escape` filters is causing it as the `link` line above doesn't seem to have this problem.

```html
<id>{{ '/' | absolute_url | xml_escape }}/</id>
```

## Expected Result

```xml
<id>http://test-site.com/</id>
```

pathawks:
This is a problem with the `absolute_url` filter in Jekyll.
jekyll/jekyll#5541

pathawks:
Hmm.

This doesn't happen in the `<link>` tag; only in the `<id>` tag. Difference is the `xml_escape` filter.

mmistakes:
@pathawks I noticed that too. Really strange.

pathawks:
There's a trailing slash _after_ the Liquid. I am banging my head against a wall right now.

https://github.com/jekyll/jekyll-feed/blob/v0.8.0/lib/feed.xml#L8

mmistakes:
Hahah. It's crazy how that extra `/` just blended in. Completely missed that!

PR #159, opened 2017-02-16
Remove trailing slash from feed ID

pathawks:
See #158

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -7,7 +7,7 @@
   <link href="{{ page.url | absolute_url }}" rel="self" type="application/atom+xml" />
   <link href="{{ '/' | absolute_url }}" rel="alternate" type="text/html" {% if site.lang %}hreflang="{{ site.lang }}" {% endif %}/>
   <updated>{{ site.time | date_to_xmlschema }}</updated>
-  <id>{{ '/' | absolute_url | xml_escape }}/</id>
+  <id>{{ '/' | absolute_url | xml_escape }}</id>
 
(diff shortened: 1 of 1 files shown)


pathawks:
Rubocop error is unrelated

benbalter:
@jekyllbot: merge +minor.

benbalter:
Nice catch @pathawks! 🌷 🐢 🌮

benbalter:
Think it's worth releasing a 0.9.0? ~35 commits on master.

pathawks:
I think @ashmaroli and I noticed it at the exact same moment.

I think it would be fine to release a 0.9.0

We should mention somewhere that _all_ feed IDs will change now. That may result in previously read items appearing as unread in feed readers. Nothing that can be done to prevent that, however.

ashmaroli:
or how about a `v1.0`? esp. with the namespace change n all.. `Jekyll::JekyllFeed` is `JekyllFeed::Generator` now.. Doesn't that qualify as a major change in API?
Please enlighten me.

Moreover, I thought removing `/` was just a bug-fix, but apparently its a `Minor Enhancement`. This became clear by reading Pat's comment above though..

pathawks:
Public API is not considered stable until `1.0.0`. According to [semver.org](http://semver.org/#spec-item-4):

> Major version zero (0.y.z) is for initial development. Anything may change at any time. The public API should not be considered stable.

We are still in the initial development stage. We will ship `1.0.0` when things are stable. We are close.

ashmaroli:
Okay. Thanks :smiley:

PR #162, opened 2017-02-17
Update feed.xml

### LINKED ISSUE #161: 0.9.0: invalid xml
```
This feed does not validate.

    line 1, column 79: XML parsing error: <unknown>:1:79: not well-formed (invalid token)
```

The problem is a missing space.

```xml
<?xml version="1.0" encoding="utf-8"?><feed xmlns="http://www.w3.org/2005/Atom"xml:lang="en">
                                                                               ^
```

XhmikosR:
Fixes #161

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -2,7 +2,7 @@
 {% if page.xsl %}
   <?xml-stylesheet type="text/xml" href="{{ '/feed.xslt.xml' | absolute_url }}"?>
 {% endif %}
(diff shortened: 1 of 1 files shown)


benbalter:
@jekyllbot: merge +minor.

benbalter:
THanks @XhmikosR!

pathawks:
I wonder if we would be better off relaxing our [`MINIFY_REGEX`](https://github.com/jekyll/jekyll-feed/blob/v0.9.0/lib/jekyll-feed/generator.rb#L19)?

XhmikosR:
That seems the proper solution. I just wanted a quick fix since the current
release was broken for me on Firefox.

Issue #163, opened 2017-02-17
Add a shell script to validate xml files with CI


ashmaroli:
We could catch bugs in the generated markup if a script was written to have `xmllint` validate the generated xml files in various situations, with Travis CI.

Thoughts?

pathawks:
Not a shell script, but we should probably use Nokogiri instead of just regex in our test suite :+1:

ashmaroli:
Nokogiri is great too.. :smiley:  :+1:

pathawks:
Wait... we [have this already](https://github.com/jekyll/jekyll-feed/blob/v0.9.1/spec/jekyll-feed_spec.rb#L100-L107)

ashmaroli:
its not a full-blown validation though..

pathawks:
It is [validated by the W3 validation service](https://github.com/jekyll/jekyll-feed/blob/v0.9.1/spec/jekyll-feed_spec.rb#L178-L206)

ashmaroli:
yet.. we still encountered a bug in the generated feed..
okay, closing this.

pathawks:
We weren't testing for cases where `site.lang` was set.

PR #164, opened 2017-02-17
Test feeds that have a `site.lang`

pathawks:
See #162 #163

### DIFF
--- a/spec/jekyll-feed_spec.rb
+++ b/spec/jekyll-feed_spec.rb
@@ -105,6 +105,7 @@ describe(JekyllFeed) do
       expect(feed.feed_version).to eql("1.0")
       expect(feed.encoding).to eql("UTF-8")
       expect(feed.lang).to be_nil
+      expect(feed.valid?).to eql(true)
     end
 
     it "outputs the link" do
@@ -144,6 +145,7 @@ describe(JekyllFeed) do
         expect(feed.feed_type).to eql("atom")
         expect(feed.feed_version).to eql("1.0")
         expect(feed.encoding).to eql("UTF-8")
+        expect(feed.valid?).to eql(true)
       end
 
       it "outputs the correct language" do
--- a/spec/jekyll-feed_spec.rb
+++ b/spec/jekyll-feed_spec.rb
@@ -156,6 +156,11 @@ describe(JekyllFeed) do
         post = feed.items.first
         expect(post.lang).to eql(lang)
       end
+
+      it "renders the feed meta" do
+        expected = %r!<link href="http://example.org/" rel="alternate" type="text/html" hreflang="#{lang}" />!
+        expect(contents).to match(expected)
+      end
     end
 
     context "with site.title set" do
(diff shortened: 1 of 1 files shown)


spec/jekyll-feed_spec.rb
```diff
@@ -267,12 +282,4 @@
       expect(contents).to include('<?xml-stylesheet type="text/xml" href="http://example.org/feed.xslt.xml"?>')
     end
   end
-
-  context "with site.lang set" do
-    let(:overrides) { { "lang" => "en-US" } }
-
-    it "should set the language" do
-      expect(contents).to match %r{type="text/html" hreflang="en-US" />}
```
pathawks: Perhaps I should leave this test, as it appears to be checking [something else](https://github.com/jekyll/jekyll-feed/blob/v0.9.1/lib/jekyll-feed/feed.xml#L8)

ashmaroli:
There's one other place in the [feed template](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L36) where things could break again..

spec/jekyll-feed_spec.rb
ashmaroli: yes, this should be left to test that parameter

pathawks: How does [this look](https://github.com/jekyll/jekyll-feed/pull/164/files#diff-46663446120a8d6231c2e8f9f752dcfcR160) as a replacement?

ashmaroli: Better.. and more complete :+1:

ashmaroli:
Because [this line](https://github.com/jekyll/jekyll-feed/blob/master/lib/jekyll-feed/feed.xml#L36) in the template could output to:
```xml
<entryxml:lang="en-us">
```
based on the recent bug, I think that should be edited on the `master` branch

ashmaroli:
```ruby
    context "with post.lang set" do
      it "renders the feed meta" do
         expected = %r!<entry xml:lang="en-us">!
         expect(contents).to match(expected)
       end
    end
```

pathawks:
Isn't that what [this](https://github.com/jekyll/jekyll-feed/pull/164/files#diff-46663446120a8d6231c2e8f9f752dcfcR157) is testing?

ashmaroli:
> Isn't that what this is testing?

does not test the entire stretch of the markup.. the missing whitespace remained undetected..

spec/jekyll-feed_spec.rb
```diff
@@ -136,6 +138,31 @@
...
+      it "outputs the correct language" do
+        expect(feed.lang).to eql(lang)
+      end
+
+      it "sets the language of entries" do
+        post = feed.items.first
+        expect(post.lang).to eql(lang)
+      end
+
+      it "renders the feed meta" do
```
pathawks: This is testing a link in the feed to the site; not the feed meta tag. Perhaps it could use a better name.

pathawks:
@jekyllbot: merge +dev

PR #165, opened 2017-03-03
Use `https` in more places.

XhmikosR:


benbalter:
@jekyllbot: merge +docs

benbalter:
Thanks @XhmikosR!

XhmikosR:
On a side note, whoever is in charge of the Cloudflare account for jekyllrb.com should enforce HTTPS with a simple Page Rule.

/CC @benbalter

PR #170, opened 2017-04-19
Update default xml for MailChimp

HaoZeke:


HaoZeke:
Any updates on this?

pup:
@HaoZeke If you could rebase `master` on top of this (e.g. updating this branch to the current state of master), I'd be more than happy to land this!

lib/jekyll-feed/feed.xml
```diff
@@ -74,7 +74,7 @@
         {% unless post_image contains "://" %}
           {% assign post_image = post_image | absolute_url | xml_escape  %}
         {% endunless %}
-        <media:thumbnail xmlns:media="http://search.yahoo.com/mrss/" url="{{ post_image }}" />
+        <media:content xmlns:media="http://search.yahoo.com/mrss/" url="{{ post_image }}" medium="image" />
```
pathawks: What is this about? Can you link to some documentation explaining this?

I don’t want to get rid of `media:thumbnail`, it’s used by feed readers.

HaoZeke:
So the current layout isn't picked up by mailchimp which expects media:content
I could also leave thumbnail in there too, good point.. @pathawks

I'll update this sometime soon.

pathawks:
You could maybe put them both in the same `media:group`?

We may have to do some experiments to see how this will be rendered in various feed readers.

DirtyF:
Fixed in #290

Issue #172, opened 2017-05-17
Plans to support JSON Feed?


reefdog:
[JSON Feed](https://jsonfeed.org) was just announced. Assuming it has any legs—Manton and Brent are no lightweights—it might stick around. Is it reasonable to expect jekyll-feed to one day support it alongside Atom, or should a new plugin be developed?

Either way, I'm keen to help develop support. Just wondering if it belongs in a new plugin or as a configurable option for this one.

pathawks:
I can't speak for @benbalter, but **I** have no plans at this point.

If there were a compelling advantage, we could take another look. I _am_ very excited about the technology, but I think we need to be conservative about what we add.

From [the JSON Feed spec](https://jsonfeed.org/version/1):

> Any publisher already publishing RSS and/or Atom should continue to do so. In fact, if you’re trying to decide which format (of RSS, Atom, and JSON Feed) to use, and you can do only one, pick RSS — it’s time-tested, popular, and good.

benbalter:
I'm with @pathawks on this one. If we add it now, we have to support it in perpetuity, even if consumers don't adopt it.

That said, we may be able to move some of the template logic to a drop like we did for SEO tag (creating shared field values), and allow another plugin to consume those calculated fields.

reefdog:
Sensible position. Thanks, both! (Feel free to close this or leave it open as a signpost, as you see fit.)

pathawks:
If we _did_ support JSON Feed, we would want to build the feed in Ruby rather than Liquid (like what SEO Tag now does for JSON-LD).

Closing for now, but if there were a compelling use case I really would love to add support. Will keep an eye on this.

pathawks:
> That said, we may be able to move some of the template logic to a drop like we did for SEO tag (creating shared field values), and allow another plugin to consume those calculated fields.

@benbalter I wonder how much shared logic there is between **jekyll-feed** and **jekyll-seo**. I wonder if it would make any sense at all to abstract some of that logic into a separate gem?

parkr:
I'm really jazzed about this project and would love to see JSON Feed supported. Could we reconsider this? Would it make more sense to ship as a separate plugin, e.g. jekyll-feed-json?

pathawks:
Maybe we 🎄 and if a site specifies a path in `site.feed.json_path` we can enable JSON Feed output?

There is enough shared logic that it would be a shame to duplicate everything in a second gem.

[...]

PR #173, opened 2017-05-17
Create JSON feed

hugmanrique:
The original spec for JSON Feeds has been published this morning ([Spec](https://jsonfeed.org/version/1)), and as my first Ruby exercise, I tried to create JSON feeds for Jekyll.

By default, the JSON feed will be written to feed.json, but a config option is also available. The JSON code should be valid, and I perform sanitization of strings by using the `json` Liquid operator.

Please note that this is the first time I've modified some Ruby code, and some things could be totally wrong (I'm not too sure about this [return](https://github.com/hugmanrique/jekyll-feed/blob/master/lib/jekyll-feed/generator.rb#L10))

### DIFF
--- a/lib/jekyll-feed/feed.json
+++ b/lib/jekyll-feed/feed.json
@@ -1,0 +1,63 @@
+{
+  "version": "https://jsonfeed.org/version/1",
+  {% assign site_title = site.title | site.name }
+  {% if site_title %}
+    "title": {{ site_title | smartify | json }},
+  {% endif %}
+  "home_page_url": "{{ '/' | absolute_url }}",
+  "feed_url": "{{ page.url | absolute_url }}",
+  {% if site.description }
+    "description": {{ site.description | json }},
+  {% endif %}
+  {% if site.author %}
+    "author": {
+      "name": {{ site.author.name | default: site.author | json }},
+      {% if site.author.uri %}
+        "url": {{ site.author.uri | json }},
+      {% endif %}
+    }
+  {% endif %}
+  "items": [
+    {% assign posts = site.posts | where_exp: "post", "post.draft != true" %}
+    {% for post in posts limmit: 10 %}
+      {
+        "id": {{ post.id | absolute_url | json }},
+        "url": "{{ post.url | absolute_url }}",
+        "title": {{ post.title | smartify | strip_html | normalize_whitespace | json }},
+        {% if post.excerpt and post.excerpt != empty %}
+          {% assign summary = post.excerpt | strip_html | normalize_whitespace | json %}
+          "content_html": {{ summary }},
+          "summary": {{ summary }},
+        {% endif %}
+        {% assign post_image = post.image.path | default: post.image %}
+        {% if post_image %}
+          {% unless post_image contains "://" %}
+            {% assign post_image = post_image | absolute_url | xml_escape  %}
+          {% endunless %}
+          "image": "{{ post_image }}",
+        {% endif %}
+        "date_published": "{{ post.date | date_to_xmlschema }}",
+        "date_modified": "{{ post.last_modified_at | default: post.date | date_to_xmlschema }}",
+        {% assign post_author = post.author | default: post.authors[0] %}
+        {% if post_author %}
+          {% assign post_author = site.data.authors[post_author] | default: post_author %}
+          {% assign post_author_uri = post_author.uri | default: nil %}
+          {% assign post_author_name = post_author.name | default: post_author %}
+          "author": {
+            {% if post_author_name %}
+              "name": {{ post_author_name | json }},
+            {% endif %}
+            {% if post_author_uri %}
+              "url": {{ post_author_uri | json }}
+            {% endif %}
+          }
+        {% endif %}
+        "tags": [
+          {% for tag in post.tags %}
+            {{ tag | json }}
+          {% endfor %}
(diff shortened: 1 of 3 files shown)


pathawks:
@humanrique This is awesome! Great work here.

Per the discussion in #172, I’m not sure that we are ready to support JSON Feed just yet. That said, I wonder if it would be worthwhile for us to have a branch that supports JSON. Kind of a [just in case](https://en.m.wikiquote.org/wiki/Steve_Jobs) operation. This way, anybody who wanted to work with JSON Feed and Jekyll could have a common place to do so.

@benbalter Thoughts?

parkr:
Huge 👍 on this.

lib/jekyll-feed/feed.json
```diff
@@ -0,0 +1,65 @@
...
+    "author": {
+      "name": {{ site.author.name | default: site.author | json }},
+      {% if site.author.uri %}
+        "url": {{ site.author.uri | json }}
+      {% endif %}
+    },
+  {% endif %}
+  "items": [
+    {% assign posts = site.posts | where_exp: "post", "post.draft != true" %}
+    {% for post in posts limmit: 10 %}
```
justinrummel: I believe "limmit" needs to be with one "m" to work correctly.

lib/jekyll-feed/generator.rb
```diff
@@ -27,11 +29,25 @@ def feed_path
...
+      if @site.config["feed"] && @site.config["feed"]["json_path"]
+        @site.config["feed"]["json_path"]
+      else
+        "feed.json"
+      end
+    end
+
     # Path to feed.xml template file
     def feed_source_path
       File.expand_path "./feed.xml", File.dirname(__FILE__)
```
bogdanvlviv:
```ruby
File.expand_path("feed.xml", __dir__)
```

lib/jekyll-feed/generator.rb
```diff
@@ -27,11 +29,25 @@ def feed_path
...
     end
 
+    # Path to feed.json template file
+    def feed_json_source_path
+      File.expand_path "./feed.json", File.dirname(__FILE__)
```
bogdanvlviv: File.expand_path("feed.xml", __dir__)

lib/jekyll-feed/feed.json
```diff
@@ -0,0 +1,77 @@
+{
+  "version": "https://jsonfeed.org/version/1",
+  {% assign site_title = site.title | site.name %}
+  {% if site_title %}
+    "title": {{ site_title | smartify | json }},
+  {% endif %}
+  {% if site.description %}
+    "description": {{ site.description | json }},
+  {% endif %}
+  "home_page_url": "{{ '/' | absolute_url }}",
```
bogdanvlviv: why just not `{{ absolute_url }}`?

pathawks: `absolute_url` is a filter not a value. It needs *something* as input.

lib/jekyll-feed/generator.rb
hugmanrique: The feed_source_path function used that same format and I just changed the file path. Please note that I don't know any Ruby and I might be wrong.

lib/jekyll-feed/feed.json
```diff
@@ -0,0 +1,77 @@
...
+        {% if post.enclosure %}
+          "attachments": [
+            {
+              "url": "{{ post.enclosure }}",
+              "mime_type": "{{ post.enclosure_type }}",
+              "size_in_bytes": "{{ post.enclosure_length }}"
+            }
+          ],
+        {% endif %}
+        "tags": [
```
bogdanvlviv:
```
"tags": {{ post.tags | jsonify }},
```

lib/jekyll-feed/feed.json
```diff
@@ -0,0 +1,72 @@
...
+  "items": [
+    {% assign posts = site.posts | where_exp: "post", "post.draft != true" %}
+    {% for post in posts limit: 10 %}
+      {
+        "id": {{ post.id | absolute_url | json }},
+        "url": "{{ post.url | absolute_url }}",
+        "title": {{ post.title | smartify | strip_html | normalize_whitespace | json }},
+        {% if post.excerpt and post.excerpt != empty %}
+          {% assign post_summary = post.excerpt | strip_html | normalize_whitespace | json %}
+          "content_html": {{ post_summary }},
```
bogdanvlviv: i think `content_html` should have a full content of post

lib/jekyll-feed/feed.json
```diff
@@ -0,0 +1,72 @@
+{
+  "version": "https://jsonfeed.org/version/1",
+  {% assign site_title = site.title | site.name %}
+  {% if site_title %}
+    "title": {{ site_title | smartify | json }},
```
bogdanvlviv: i can't find `json` filter there https://jekyllrb.com/docs/templates/ may be `jsonify`

pathawks:
This is a cool proof of concept but honestly, if we were to produce a JSON feed, I think doing it in pure Ruby would be a better approach then using Liquid templates.

hugmanrique:
The XML feed is currently also produced through a Jekyll template, and I believe being able to see how the template is going to end up looking before executing the script is better than creating an object through Ruby.

Issue #70, opened 2015-11-02
Option to have feeds by tags/categories


nasht00:
I would like to have a feed such as /blog/feed/MyGreatTag

parkr:
@benbalter, any thoughts here? I have no thoughts other than "seems neat, but is it core to this plugin?"

benbalter:
Been thinking about this a lot lately (category- or tag-specific views). 

On the one hand, this is a staple of how many blogs have functioned for the past ten years. The only reason Jekyll blogs traditionally don't do this (or e.g., have the full content of the post on the index page, or paginate) is because Jekyll doesn't support that easily. Instead, most Jekyll blogs don't paginate, don't have category indexes, and simply list the titles/dates of posts when in list views. It's a technical constraint influencing behavior.

On the other hand, one could argue the only reason we've been trained to do that, is because it's been a prevalent feature in WordPress for about 10 years, and as developers, we're trained to **never ever, ever** remove a feature. Since it exists in WordPress, and thus exists in most blogs, it must exist in Jekyll.

I've been struggling with this looking at GitHub's own blog, which is currently not Jekyll, and has category archives (and I believe category-specific RSS feeds). Looking at actual usage, it's negligible (less than 1%). The same is true for pagination (basically nobody clicks into the 2nd page for any blog view). 

It'd be interesting to make the qualitative case for where such features are used in practice, rather than the philosophical case for why these features should exist in theory. I'm torn however, because if a developer does a bake off between WordPress and Jekyll, absent significant education, WordPress is going to "win" each time. 

It's almost as if we have to support the feature, because it's always been a feature of blogs, not because anyone uses it.

All that said, @nasht00, what's your use case? (and are you using any other index/archive plugins that might be better suited to implement the functionality via an API?)

nasht00:
Thanks @benbalter for truly considering :)

My use-case is exactly the one you covered. We are migrating from Wordpress, and people are subscribed to tag-specific RSS feeds today. But also I think it's a nice feature in any case. For technical blogs that cover various topics, it's nice that you can subscribe to a specific topic.

Since the time I posted this question here, things I've changed. I realized that restricting myself to GitHub-approved plugins only was becoming harder, so I switched to offline (Travis)  build (and upload the generated HTML).
With this I was able to switch to Jekyll 3 and start writing my own custom plugins.

This what my plugin looks like so far:

```ruby
module Jekyll
  # This generates a Page showing all blog posts for a specific tag
[...]

PR #176, opened 2017-05-30
Generate a feed for each category.

mortenson:
Like others in #70, I want to have a feed per-category so that readers can subscribe only to topics that interest them. When testing this PR, you should be able to visit `/feed/category.xml`, where "category" is your category name.

I've only implemented categories here, as it seems like there would be loads of tags and users may expect complicated syntax like: `/feed/tag1/tag2/combined.xml`. Categories are how you categorize your content, so it makes sense that they would represent individual feeds.

I just started using Jekyll, and have never written Ruby/Liquid before, so forgive me if anything is amiss here.

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -33,6 +33,11 @@
 
   {% assign posts = site.posts | where_exp: "post", "post.draft != true" %}
   {% for post in posts limit: 10 %}
+    {% if page.category %}
+      {% unless post.categories contains page.category %}
+        {% continue %}
+      {% endunless %}
+    {% endif %}
     <entry{% if post.lang %}{{" "}}xml:lang="{{ post.lang }}"{% endif %}>
       <title type="html">{{ post.title | smartify | strip_html | normalize_whitespace | xml_escape }}</title>
       <link href="{{ post.url | absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -6,8 +6,15 @@ module JekyllFeed
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
-      return if file_exists?(feed_path)
-      @site.pages << content_for_file(feed_path, feed_source_path)
+      unless file_exists?(feed_path)
+        @site.pages << content_for_file(feed_path, feed_source_path)
+      end
+      @site.categories.each do |category|
+        path = "/feed/" + category.first + ".xml"
+        unless file_exists?(path)
+          @site.pages << content_for_file(path, feed_source_path, category.first)
+        end
+      end
     end
 
     private
@@ -42,12 +49,13 @@ module JekyllFeed
     end
 
     # Generates contents for a file
-    def content_for_file(file_path, file_source_path)
+    def content_for_file(file_path, file_source_path, category = false)
       file = PageWithoutAFile.new(@site, File.dirname(__FILE__), "", file_path)
       file.content = File.read(file_source_path).gsub(MINIFY_REGEX, "")
       file.data["layout"] = nil
       file.data["sitemap"] = false
       file.data["xsl"] = file_exists?("feed.xslt.xml")
+      file.data["category"] = category
       file.output
       file
     end


benbalter:
@mortenson this great. Thanks for opening the PR.

I'd be curious your thoughts on https://github.com/jekyll/jekyll-feed/issues/70#issuecomment-159442173 and the subsequent discussion. It sounds like you're moving from a platform that offers category feeds and are seeking to maintain feature parity. If we add a feature, we have to support it in perpetuity, regardless of how often it's used. Do you have any stats on how often category feeds are subscribed to?

mortenson:
I don't have much quantitative data - but can speak to why I would like to see this feature. As I mentioned earlier, I just started using Jekyll this week, and got a minimal blog up at http://mortenson.coffee/blog. This is my personal site, and I plan on writing posts about a variety of topics.

The problem I'm running into is that I'd like to add my blog to [Planet Drupal](https://www.drupal.org/planet), which is an aggregate of a ton of Drupal community blogs, but they specifically require that your RSS feed only contains Drupal-related content:

> You must have an RSS feed that contains only content appropriate for Drupal Planet. If your site uses Drupal, you can create a "Drupal Planet" taxonomy term and use it to tag any content you want to appear on the Planet. Drupal automatically creates an RSS feed for each taxonomy term, and you can submit that term’s feed URL in your Drupal Planet application. Alternatively, you can use the Drupal Planet feature module, which uses Flag and Views modules to create the feed. If you are not using Drupal, you will need to make sure your site can create a feed just for your Planet content.
[source](https://www.drupal.org/drupalorg/docs/content/planet-drupal)

My blog at this point would not qualify for aggregation here, as I plan to mix personal posts with Drupal-related posts. There are 615 sites in this aggregate, and each site provides a specific RSS feed just for Drupal. It helps that most of those blogs _are also built with Drupal_, which provides this functionality out of the box and is highly utilized. A similar CMS, Wordpress, lets you filter arbitrarily by multiple criteria (author, tag, date, etc.), which obviously isn't a fit for static sites.

In the short-term, I could just create a hard-coded RSS page in Jekyll that only shows Drupal posts, but if I want to add my blog to a security aggregate I would have to do the same task. The same goes for displaying just my personal posts on another site - without this functionality I'll end up maintaining my own RSS feeds, which means I wouldn't be using jekyll-feed at all.

RobbiNespu:
I often use category feeds are subscribed to Fedora planet which which is an aggregate feed of Fedora communities. I am using blogspot before, now after move to jekyll.. I need this features on my theme which  using jekyll-feed plugin.

marco-c:
Same on Planet Mozilla.

halirutan:
Same for user feeds on specific StackExchange chats. Although my community is very open and wouldn't bother some off-topic posts, I would like to separate this too. Not that I have time to write many posts, but this is another issue.

kylebarbour:
I would benefit from this substantially as my site currently has several separate sections with significantly disparate content &mdash; few, if any of my subscribers use both. Migrating to Jekyll has meant that new posts are substantially easier, but a unified feed for everything on the site doesn't meet my needs. This particular feature seems to me to be a fairly minimal change and one worth supporting.

If the concern is that automatic generation of new feeds is unwanted &mdash; as that changes the default output &mdash; a simple solution would be to enable specific categories in your `_config.yml`. In other words, for someone with the categories `xyzzy`, `plugh`, and `thud`, the default would still be to generate a single unified `feed.xml`. However, with:

```yml
feed:
    multifeed:
        - xyzzy
        - plugh
```

in the user's `_config.yml`, the unified `feed.xml` would still be generated, as would the category-specific `xyzzy.xml` and `plugh.xml` feeds (but not `thud`, which was not specified).

Thoughts?

benbalter:
I think I was the one that originally :-1:'d this feature. I still stand by my comments in https://github.com/jekyll/jekyll-feed/issues/70#issuecomment-159442173, but it's clear that it's something many users want to implement (even if what little data we have suggests it's not actually used). If we can implement this in a way that minimizes complexity and the burden on the user, I'd be :+1: to supporting this going forward (and will leave some comments on the PR about the code itself).

lib/jekyll-feed/feed.xml
```diff
@@ -33,6 +33,11 @@
 
   {% assign posts = site.posts | where_exp: "post", "post.draft != true" %}
   {% for post in posts limit: 10 %}
+    {% if page.category %}
```
benbalter: We'll need to filter the category above, otherwise the limit will be off. I believe the following should work:

`site.categories[page.category]  | where_exp: "post", "post.draft != true"`

lib/jekyll-feed/generator.rb
```diff
@@ -8,8 +8,15 @@ class Generator < Jekyll::Generator
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
-      return if file_exists?(feed_path)
-      @site.pages << content_for_file(feed_path, feed_source_path)
+      unless file_exists?(feed_path)
+        @site.pages << content_for_file(feed_path, feed_source_path)
+      end
+      @site.categories.each do |category|
+        path = "/feed/" + category.first + ".xml"
```
benbalter: What is `category.first`?

lib/jekyll-feed/generator.rb
```diff
@@ -44,12 +51,14 @@ def file_exists?(file_path)
...
     # Generates contents for a file
-    def content_for_file(file_path, file_source_path)
+
+    def content_for_file(file_path, file_source_path, category = false)
       file = PageWithoutAFile.new(@site, __dir__, "", file_path)
       file.content = File.read(file_source_path).gsub(MINIFY_REGEX, "")
       file.data["layout"] = nil
       file.data["sitemap"] = false
       file.data["xsl"] = file_exists?("feed.xslt.xml")
+      file.data["category"] = category
```
benbalter: Would it be cleaner to allow the template to accept an arbitrary array of documents? That way, we can pass it all post, a category, a tag, or even another collection.

halirutan:
@benbalter Thanks for pushing this PR forward. This is highly appreciated.

halirutan:
@mortenson Any chance you can fix the issues pointed out by @benbalter? I'm still keen to see this merged. Or has a similar feature been added to Jekyll in the meantime by a different PR?

mortenson:
@halirutan I believe #228 is the PR to watch now - @benbalter is working on it.

mortenson:
#228 Has been merged, closing.

PR #187, opened 2017-08-24
Define path with __dir

DirtyF:
Inspired from https://github.com/jekyll/jekyll/pull/6087 by @bogdanvlviv


parkr:
@jekyllbot: merge

PR #188, opened 2017-08-25
Bump Ruby for Travis

DirtyF:
Use latest Ruby like GitHub Pages


DirtyF:
@pathawks Right, updated to be coherent with other plugins.

benbalter:
@jekyllbot: merge.

PR #193, opened 2017-10-08
Make feed limit configurable

### LINKED ISSUE #192: Feed Limit?
Is there an option to specify the number of entries the feed should contain? I'm looking to generate a feed with 50 entries but currently, it only has a few listed.

Thanks,
Ben

DirtyF:
fix #192 #56


pathawks:
Rather than changing the `limit`, I’d rather revisit trying to include the union of *most recent 10 posts* and *posts from the last 48 hours*.

I worry that one more config option without an obvious best setting is problematic.

DirtyF:
@pathawks That's an interesting alternative approach that could feel like magic ✨  even if that behavior makes sense for websites who publish a lot of content. Where do you get the 48h limit from? Is that a good practice?

I went the (simple) explicit way and let people decide what _they_ wanna do. 
Most users shouldn't have to change the default anyway, we can all agree to that.

pathawks:
> Where do you get the 48h limit from? Is that a good practice?

I can’t remember where that number came from, but the argument to increase the `limit` comes from the problem of a site publishing new items faster than consumers are checking for updates. A limit of 48 hours ensures that consumers will get all new items, enen if they only poll every other day.

pathawks:
> Most users shouldn't have to change the default anyway, we can all agree to that.

Yes.

I would add that most users should never think about the contents of their feed XML file. If this plugin does its job, things should Just Work.

pathawks:
> The union of the last 48 hours and 10 posts doesn't work for me as I update with long delays with interspersed frequent updates, so that would either create an empty feed (when I don't post for more than 48 hours) or the possibility infrequent pollers of missing updates during the high-intensity periods.

As long as the feed is polled every 2 days, no updates would be missed.

PR #197, opened 2017-11-03
Fix: Add note about using plugins instead of gems key

localheinz:
This PR

* [x] updates `README.md` with a note about using the `plugins` instead of the `gems` key for enabling the plugin

💁‍♂️ For reference, see https://github.com/jekyll/jekyll/pull/5130.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,8 @@ gems:
   - jekyll-feed
 ```
 
+:bulb: If you are using Jekyll 3.5.0 or greater, you probably want to use the `plugins` key instead of `gems`.
+
 ## Usage
 
 The plugin will automatically generate an Atom feed at `/feed.xml`.
--- a/README.md
+++ b/README.md
@@ -15,11 +15,11 @@ gem 'jekyll-feed'
 And then add this line to your site's `_config.yml`:
 
 ```yml
-gems:
+plugins:
   - jekyll-feed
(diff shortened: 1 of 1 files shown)


DirtyF:
While we should encourage the use of this config key, we must not forget those using older versions of Jekyll. This works only for Jekyll > 3.5.0, and will display a possibly confusing error message as `plugins` is the former name of `plugins_dir` config key.

localheinz:
@DirtyF 

Good point! Would you prefer me to close this PR, or adjust the instructions so they reflect the proper usage on older and new versions of Jekyll?

DirtyF:
@localheinz Please do upgrade the instructions ✍️

localheinz:
@DirtyF 

👍

README.md
```diff
@@ -19,6 +19,8 @@ gems:
   - jekyll-feed
 ```
 
+:bulb: If you are using Jekyll `^3.5.0`, you probably want to use the `plugins` key instead of `gems`.
```
DirtyF: I'm not sure everybody understand this notation. What about plain text?

If you are using Jekyll version `3.5.0` or greater…

DirtyF:
@jekyllbot: merge +docs

localheinz:
Thank you, @DirtyF!

PR #205, opened 2018-01-16
Add documentation for disabling smartify filter

kylebarbour:
This small PR simply explains how to disable the smartify filter. I spent the better part of the evening trying to figure this out so thought I'd write up the solution to hopefully help others as well.

Hope it's useful!


DirtyF:
@jekyllbot: merge +docs

DirtyF:
Thanks @kylebarbour, much appreciated 😄

pathawks:
Purely out of curiosity, what is your use-case for disabeling SmartyPants?

kylebarbour:
Thanks all!

@pathawks: almost purely aesthetic I'm afraid. I got annoyed by so-called smart punctuation a long time ago when copy/pasting from materials with them to things that expect ASCII input and in the process trained myself to prefer the neutral ASCII quotes. Maybe I'm just a curmudgeon. Maybe there's a way to display them as smart but have them copy/pasted as neutral, but I haven't figured it out if so :)

Issue #225, opened 2018-05-06
multiple feed templates


nebulousdog:
is there a good tactic to support multiple feed templates? i.e. if I want to create two templates and have them both generated every time, where should I tie that in?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

nebulousdog:
related PR: https://github.com/jekyll/jekyll-feed/pull/227

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

nebulousdog:
hoping to be able to refresh my PR this week, sorry for the delay. thanks @jekyllbot.

PR #227, opened 2018-05-07
adds multiple feed template support

dreamalligator:
this PR adds support for multiple feed templates. it is kindof like https://github.com/jekyll/jekyll-feed/pull/176, but this allows any number of custom templates, not just category support.

it preserves the previous feed-path config options as well as a new format 

```
feed:
  paths:
    - feed_source_path: feed.xml # default feed template
      feed_output_path: feed.xml
    - feed_source_path: _layouts/special-category-feed.xml # example
      feed_output_path: very-special-feed.xml
```

closes https://github.com/jekyll/jekyll-feed/issues/225.

### DIFF
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
+DEFAULT_ENTRIES = [{ "feed_source_path" => "feed.xml", "feed_output_path" => "feed.xml" }]
+
 module JekyllFeed
   class Generator < Jekyll::Generator
     safe true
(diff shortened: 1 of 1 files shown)


benbalter:
@nebulousdog Can you share a bit more about what problem you're trying to solve here? What's the advantage of this approach versus just putting additional templates in the site source and having Jekyll build them normally?

nebulousdog:
sure! I just saw your PR  and that's a great feature (https://github.com/jekyll/jekyll-feed/pull/228). 

I have two use-cases that are slightly different, because they aren't collection/category based.

In the first use-case; I'm using a second "mini" feed that is a stripped down version of the default `feed.xml`. Each generation creates the default feed for users to subscribe to. The "mini" feed is used on the parent site to show links to the blog. This is slimmed down so that the ajax response is as small as possible, and isn't intended to be subscribed to.

The second use-case is for archiving by language;  I have some archives based on locale/language (the `lang` frontmatter attribute `jekyll-feed`). Since these aren't based on categories/tags, they have a similar custom template requirement.

Thanks!

PR #228, opened 2018-05-08
Categories and collections

### LINKED ISSUE #70: Option to have feeds by tags/categories

benbalter:
This pull request builds on @mortenson's work in https://github.com/jekyll/jekyll-feed/pull/176 to fix #70 by abstracting some of the existing logic out a bit. Specifically, it supports generating feeds for categories (and collections other than posts). 

To generate feeds for categories:

```yml
feed:
  categories:
    - news
    - updates
```

And to generate feeds for collections:

```yml
feed:
  collections:
    - changes
```

Since posts are a collection, it's also possible to generate category feeds for other collections:

```yml
feed:
  collections:
    changes:
      categories:
        - news
        - updates
```

Everything is backwards compatible with the existing configuration flags, such that:

```yml
feed:
  path: "foo.xml"
```

Automatically gets converted to:

```yml
feed:
  collections:
    posts: 
      path: "foo.xml"
```

So this should be a non-breaking change, and theoretically add as little complexity as necessary to the publishing experience (in most cases, just creating a top-level list in the config to make it opt-in).

By default, categories are in the form of `/feed/<category>.xml`, collections are `/feed/<collection>.xml`, and collection categories are `/feed/<collection>/<category>/xml`.

All this is implemented with a small abstraction in the generator, which passes the collection (defaulting to posts) and optionally the category to the existing template.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,48 @@ Want to style what your feed looks like in the browser? Simply add an XSLT at `/
 
 Great question. In short, Atom is a better format. Think of it like RSS 3.0. For more information, see [this discussion on why we chose Atom over RSS 2.0](https://github.com/jekyll/jekyll-rss-feed/issues/2).
 
+## Categories
+
+Jekyll Feed can generate feeds for each category. Simply define which categories you'd like feeds for in your config:
+
+```yml
+feed:
+  categories:
+    - news
+    - updates
+```
+
+## Collections
+
+Jekyll Feed can generate feeds for collections other than the Posts collection. This works best for chronological collections (e.g., collections with dates in the filenames). Simply define which collections you'd like feeds for in your config:
+
+```yml
+feed:
+  collections:
+    - changes
+```
+
+By default, collection feeds will be outputted to `/feed/<COLLECTION>.xml`. If you'd like to customize the output path, specify a collection's custom path as follows:
+
+```yml
+feed:
+  collections:
+    changes:
+      path: "/changes.xml"
+```
+
+Finally, collections can also have category feeds which are outputted as `/feed/<COLLECTION>/<CATEGORY>.xml`. Specify categories like so:
+
+```yml
+feed:
+  collections:
+    changes:
+      path: "/changes.xml"
+      categories:
+        - news
+        - updates
+```
+
 ## Contributing
 
 1. Fork it (https://github.com/jekyll/jekyll-feed/fork)
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -8,8 +8,15 @@ module JekyllFeed
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
-      return if file_exists?(feed_path)
-      @site.pages << content_for_file(feed_path, feed_source_path)
+      unless file_exists?(feed_path)
+        @site.pages << content_for_file(feed_path, feed_source_path)
+      end
+      @site.categories.each do |category|
+        path = "/feed/" + category.first + ".xml"
+        unless file_exists?(path)
+          @site.pages << content_for_file(path, feed_source_path, category.first)
+        end
+      end
     end
 
     private
@@ -44,12 +51,14 @@ module JekyllFeed
     end
 
     # Generates contents for a file
-    def content_for_file(file_path, file_source_path)
+
+    def content_for_file(file_path, file_source_path, category = false)
       file = PageWithoutAFile.new(@site, __dir__, "", file_path)
       file.content = File.read(file_source_path).gsub(MINIFY_REGEX, "")
       file.data["layout"] = nil
       file.data["sitemap"] = false
       file.data["xsl"] = file_exists?("feed.xslt.xml")
+      file.data["category"] = category
       file.output
       file
     end
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -36,12 +36,12 @@ module JekyllFeed
     # collection - the name of a collection, e.g., "posts"
     # category - a category within that collection, e.g., "news"
     #
-    # Will return "feed.xml", or the config-specified default feed for posts
+    # Will return "/feed.xml", or the config-specified default feed for posts
     # Will return `/feed/category.xml` for post categories
     # WIll return `/feed/collection.xml` for other collections
     # Will return `/feed/collection/category.xml` for other collection categories
     def feed_path(collection: "posts", category: nil)
-      prefix = collection == "posts" ? "/feed" : "feed/#{collection}"
+      prefix = collection == "posts" ? "/feed" : "/feed/#{collection}"
       if category
         "#{prefix}/#{category}.xml"
       elsif collections[collection] && collections[collection]["path"]
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -44,7 +44,7 @@ module JekyllFeed
       prefix = collection == "posts" ? "/feed" : "feed/#{collection}"
       if category
         "#{prefix}/#{category}.xml"
-      elsif collections[collection]["path"]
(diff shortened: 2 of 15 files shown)


mortenson:
Thanks for pushing this forward @benbalter !

lib/jekyll-feed/generator.rb
```diff
@@ -20,18 +26,59 @@ def generate(site)
...
+    #
+    # Will return "feed.xml", or the config-specified default feed for posts
+    # Will return `/feed/category.xml` for post categories
+    # WIll return `/feed/collection.xml` for other collections
+    # Will return `/feed/collection/category.xml` for other collection categories
+    def feed_path(collection: "posts", category: nil)
+      prefix = collection == "posts" ? "/feed" : "feed/#{collection}"
+      if category
+        "#{prefix}/#{category}.xml"
+      elsif collections[collection]["path"]
```
parkr: We should do nil checking here in case `collections` is nil or `collections[collection]` is nil

benbalter:
> Too bad we can't incorporate collection settings with feed settings, since we're basically duplicating that

The settings for this were the toughest part. We could do, e.g.,:

```yml
collections:
  puppies:
    feed: true
```

or 

```yml
collections:
  puppies:
    feed: 
      path: foo.xml
      categories:
        - foo
```

But that felt more complex and it would burry feed settings in collection settings, rather than following the pattern of a top-level key named after the plugin.

Could see it either way (and could always do both, although that feels SUPER complicated to explain the three ways you could configure it). 

Unless anyone is strongly opinionated about the config, it sounds like this can merge with a bit more time for feedback?

pathawks:
I would like to review this. Busy at the moment; please just give me a day or so 🙂

lib/jekyll-feed/generator.rb
```diff
@@ -20,18 +26,59 @@ def generate(site)
...
+    #
+    # collection - the name of a collection, e.g., "posts"
+    # category - a category within that collection, e.g., "news"
+    #
+    # Will return "feed.xml", or the config-specified default feed for posts
+    # Will return `/feed/category.xml` for post categories
+    # WIll return `/feed/collection.xml` for other collections
+    # Will return `/feed/collection/category.xml` for other collection categories
+    def feed_path(collection: "posts", category: nil)
+      prefix = collection == "posts" ? "/feed" : "feed/#{collection}"
```
ashmaroli: why is there a leading slash for `/feed.xml` but not for `feed/#{collection}.xml` .. also, the comments above actually states the opposite..

spec/jekyll-feed_spec.rb
```diff
@@ -239,9 +239,9 @@
     end
 
     it "correctly adds the baseurl to the posts" do
-      expect(contents).to match /http:\/\/example\.org\/bass\/2014\/03\/04\/march-the-fourth\.html/
-      expect(contents).to match /http:\/\/example\.org\/bass\/2014\/03\/02\/march-the-second\.html/
-      expect(contents).to match /http:\/\/example\.org\/bass\/2013\/12\/12\/dec-the-second\.html/
+      expect(contents).to match /http:\/\/example\.org\/bass\/updates\/2014\/03\/04\/march-the-fourth\.html/
```
ashmaroli: not related to this PR.. but IMO enforcing the `percent_r` literal on the test files will avoid the need to escape the slashes here..

/cc @DirtyF

lib/jekyll-feed/generator.rb
```diff
@@ -8,8 +8,14 @@ class Generator < Jekyll::Generator
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
-      return if file_exists?(feed_path)
-      @site.pages << content_for_file(feed_path, feed_source_path)
+      collections.each do |name, meta|
+        Jekyll.logger.info "Jekyll Feed:", "Generating feed for #{name}"
+        (meta["categories"] + [nil]).each do |category|
```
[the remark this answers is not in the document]
benbalter: Passing, e.g., `collection: "posts", category: nil` will generate the main/non-category feed for that collection. It was the only way I could think of to avoid needing to write the loop twice (since the logic is the same, just without the filter).

lib/jekyll-feed/generator.rb
benbalter: Good catch. That is an oversight. Will push up a fix.

spec/jekyll-feed_spec.rb
benbalter: Yeah. Given the time, I'd also love to add unit tests to the generator and meta tag (rather than just relying on integration tests).

halirutan:
That is great! When can we expect this getting merged?

README.md
```diff
@@ -140,6 +140,48 @@ Want to style what your feed looks like in the browser? Simply add an XSLT at `/
...
+
+Jekyll Feed can generate feeds for collections other than the Posts collection. This works best for chronological collections (e.g., collections with dates in the filenames). Simply define which collections you'd like feeds for in your config:
+
+```yml
+feed:
+  collections:
+    - changes
+```
+
+By default, collection feeds will be outputted to `/feed/<COLLECTION>.xml`. If you'd like to customize the output path, specify a collection's custom path as follows:
```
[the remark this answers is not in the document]
halirutan: Or active voice: "The default output-path for collection feeds is `/feed/<COLLECTION>.xml`"

mattmcmanus:
I hate to be annoying, but I would LOVE to be able to use this feature. It looks like this PR has several approvals. Who's responsible to handle the merge and release? Is there anything a non-maintainer can do to help out here?

DirtyF:
@jekyllbot: :shipit: +minor

DirtyF:
Thanks @mattmcmanus for the reminder, feel free to test the current master branch if you wanna test the feature before we release a proper gem

mattmcmanus:
Yeah!

DirtyF:
works like a charm, expect a release soon: https://frank.taillandier.me/feed/agile.xml

halirutan:
I second this. A big YEAH!

meme

dpocock:
I've tried to use this on a Fedora system and it doesn't appear to be working:

```
$ rpm -qa | grep jekyll
rubygem-jekyll-feed-0.12.1-1.fc29.noarch
rubygem-jekyll-3.8.5-1.fc29.noarch
rubygem-jekyll-watch-2.2.1-1.fc29.noarch
rubygem-jekyll-sass-converter-1.5.2-2.fc29.noarch
rubygem-jekyll-seo-tag-2.6.0-1.fc29.noarch
```


I tried all the different permutations for `category` and `categories`

Here is a snippet of my `_config.yml`

```
plugins_dir:
  - jekyll-feed

feed:
  categories:
    - fedora
...
```

and from one of the blogs `_posts/2019-04-07-sfk-2019-prishtina-toastmasters.html`

```
...
category: fedora
...
```

and I also tried:
```
...
category: ['fedora']
...
```

A category-specific feed XML simply doesn't appear:

```
$ jekyll build
$ !fi
find . -name '*.xml'
./_site/feed.xml
./feed.xml
```

How should somebody troubleshoot in a situation like this?

ellenwyllie:
This isn't working for me either. I added 
```
feed:
  categories:
    - jakartaee
```
to my _config.yml and `category: jakartaee` to my blog post's front matter. 

The feed for the category is being created, but it doesn't have any posts in it.

I've tried variations of category in the post, such as `category: ['jakartaee']`, `categories: jakartaee` and `categories: ['jakartaee']` but none of them seem to work.

Does anyone know why this isn't working?

DirtyF:
@ellenwyllie : `category: jakartaee` should work.

steven1046:
Does jekyll-feed rely on a certain version of jekyll-assets to work properly?

mortenson:
@DirtyF `categories:` is what's documented in README.md, so I would expect it to work.

marco-c:
See also #246.

PR #233, opened 2018-06-08
Dont forget about categories

kenman345:
Since the frontmatter for posts can use `category` OR `categories` as per [documentation from jekyllrb.com](https://jekyllrb.com/docs/frontmatter/#predefined-variables-for-posts), it would make sense to also allow for categories to populate the XML data.

This change does not appear to be of a breaking type, as we already had the potential of multiple category tags from the combined use of `category` and `tags`. This just opens up the potential for even more `category` tags to be in the feed.xml thats output

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -59,6 +59,10 @@
 
       {% if post.category %}
         <category term="{{ post.category | xml_escape }}" />
+      {% elsif post.categories %}
+        {% for category in post.categories %}
+          <category term="{{ category | xml_escape }}" />
+        {% endfor %}
       {% endif %}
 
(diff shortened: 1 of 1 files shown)


DirtyF:
As long as we follow the spec, I'm 👍 

>  You may include as many category elements as you need to…

https://cyber.harvard.edu/rss/rss.html#ltcategorygtSubelementOfLtitemgt

hozza:
While we are at it with building feeds for tags in #264 - using `categories` as well as `categories` from front-matter should already be the case IMO! #264 This one is ready to merge too, lets make feeds a really useable Jekyll feature!

DirtyF:
@jekyll: merge +minor

PR #234, opened 2018-07-05
Remove check for older version of Jekyll

ashmaroli:
Since the plugin requires `~> Jekyll 3.3`, the check for whether `site` responds to `in_source_dir` (available since Jekyll-3.0) is unnecessary.

### DIFF
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -36,11 +36,7 @@ module JekyllFeed
 
     # Checks if a file already exists in the site source
     def file_exists?(file_path)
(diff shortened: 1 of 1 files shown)


pathawks:
https://github.com/jekyll/jekyll-feed/blob/bdd6f1b1ee4d5f855c9270e4fc724f5fb11b4d2b/jekyll-feed.gemspec#L23

https://github.com/jekyll/jekyll/blob/v3.3.0/lib/jekyll/site.rb#L361-L371

:+1:

DirtyF:
@jekyllbot: merge +minor

PR #235, opened 2018-07-05
Memoize computing paths to feed template and file

ashmaroli:
Since the generator runs during every regeneration cycle and even during incremental builds, its better to cache the paths to the feed-template and the initialized object

### DIFF
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -22,16 +22,18 @@ module JekyllFeed
 
     # Path to feed from config, or feed.xml for default
     def feed_path
-      if @site.config["feed"] && @site.config["feed"]["path"]
-        @site.config["feed"]["path"]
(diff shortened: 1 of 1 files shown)


lib/jekyll-feed/generator.rb
```diff
@@ -22,16 +22,18 @@ def generate(site)
...
+        else
+          "feed.xml"
+        end
       end
     end
 
     # Path to feed.xml template file
     def feed_source_path
-      File.expand_path "feed.xml", __dir__
+      @feed_source_path ||= File.expand_path("feed.xml", __dir__)
```
pathawks: Don't we only ever need this once?

ashmaroli:
Just realized this has already been proposed as part of #228 .. My bad..

DirtyF:
Closing then 😃

PR #236, opened 2018-07-09
Implement RFC5005, "Feed Paging and Archiving".

jameysharp:
This standard, published in 2007, allows feed readers to efficiently
page through the entire history of a feed, rather than being limited to
the most recent 10 posts.

This implementation supports RFC5005 section 2, "Complete Feeds", for
sites with 10 or fewer posts; and section 4, "Archived Feeds", for sites
with more than 10 posts.

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -4,8 +4,20 @@
 {% endif %}
 <feed xmlns="http://www.w3.org/2005/Atom" {% if site.lang %}xml:lang="{{ site.lang }}"{% endif %}>
   <generator uri="https://jekyllrb.com/" version="{{ jekyll.version }}">Jekyll</generator>
+
+  {% unless page.current or page.prev_archive %}
+    <fh:complete xmlns:fh="http://purl.org/syndication/history/1.0"/>
+  {% endunless %}
+  {% if page.current %}
+    <fh:archive xmlns:fh="http://purl.org/syndication/history/1.0"/>
+    <link href="{{ page.current.url | absolute_url }}" rel="current"/>
+  {% endif %}
+  {% if page.prev_archive %}
+    <link href="{{ page.prev_archive.url | absolute_url }}" rel="prev-archive"/>
+  {% endif %}
   <link href="{{ page.url | absolute_url }}" rel="self" type="application/atom+xml" />
   <link href="{{ '/' | absolute_url }}" rel="alternate" type="text/html" {% if site.lang %}hreflang="{{ site.lang }}" {% endif %}/>
+
   <updated>{{ site.time | date_to_xmlschema }}</updated>
   <id>{{ '/' | absolute_url | xml_escape }}</id>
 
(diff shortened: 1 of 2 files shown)


jameysharp:
I haven't written any significant Ruby before, so I guess I'm not surprised that CI reports there are a bunch of code style issues. I'll have to figure out how to use Rubocop after I get some sleep, I guess. Meanwhile, I'd love to hear what you think of this pull request!

I have a copy of this code running on my personal blog so you can find the feeds it generated from <https://jamey.thesharps.us/feed.xml>.

benbalter:
@jameysharp do you have any sense of adoption of the spec? E.g., do any major readers respect/use it? A quickly google search led me to https://github.com/samuelclay/NewsBlur/issues/1109, which a few weeks ago you suggested NewsBlur would be an "early adopter"? 

My concern based on the diff is that this adds a lot of complexity, and I'd hate to make the code harder to maintain for a feature no or few readers will use.

jameysharp:
Sure, I can understand that. The problem I'm trying to tackle is there's a catch-22: if publishers don't implement it, readers won't implement it, and vice versa.

I heard from Manton at micro.blog that a Jekyll implementation of this standard would solve a problem he has with podcasts, and I've heard about the same podcast problem from quite a few people. Podcast publishers currently have to choose between listing all their episodes in one feed so iTunes sees them all, or using a smaller feed to avoid making other feed readers choke. If we can drive adoption of this standard, they won't have to choose.

My personal interest is for webcomics and serial stories, which I built https://www.comic-rocket.com as a special-purpose reader for. I fully expect a new wave of Comic Rocket-like tools to appear if people start offering full-history feeds.

This is the third publisher-side implementation I've written in the last few weeks, so naturally it doesn't seem very complicated to me. Maybe it can be refactored so it doesn't seem complicated to you either? I certainly sympathize with your concern about maintenance burden.

pathawks:
> I heard from Manton at micro.blog that a Jekyll implementation of this standard would solve a problem he has with podcasts

This plugin does not create a podcast feed, so iTunes isn't really part of the equation here. I would be very curious to see a list of feed readers that would support this feature.

pathawks:
With 004bd3bf93114462a233d0faab7c999518496280 I have fixed _most_ of Rubocop's complaints. ✌️

The last remaining issue is that `generate` is too complicated. Some of the logic will need to be extracted into a helper method.

jameysharp:
Thanks for the Rubocop fixes! I've refactored further to eliminate its complaint about the AbcSize metric, and now there are no Rubocop complaints. Hooray! Is this better?

samuelclay:
> @jameysharp do you have any sense of adoption of the spec? E.g., do any major readers respect/use it? A quickly google search led me to [samuelclay/NewsBlur#1109](https://github.com/samuelclay/NewsBlur/issues/1109), which a few weeks ago you suggested NewsBlur would be an "early adopter"?
> 
> My concern based on the diff is that this adds a lot of complexity, and I'd hate to make the code harder to maintain for a feature no or few readers will use.

Funny to see this thread linked here. NewsBlur is about to launch RSS back fill for premium archive subscribers, and the NewsBlur blog recently moved to Jekyll. So I went to check if Jekyll supports back fill and it seems it will as soon as this PR is merged. Anything I can do to help prioritize that? Happy to blog about Jekyll supporting RSS pagingation, joining WordPress in automatically enabling feed paging. 

The impetus for bringing complete archives to NewsBlur is how misinformation and censorship are changing the way we consume news, and a searchable archive pushes back on the effort to rewrite the past. We need access to news and information published in the past because we need to be concerned about how our own present is remembered in the future. Having historical context, in the same interface and reader as today's news, buys optionality for those finding themselves in oppressive societies.

Thanks @jameysharp for bringing this together!

benbalter:
>  do you have any sense of adoption of the spec? E.g., do any major readers respect/use it? 

@samuelclay thanks for reviving this old thread. Given that it's been nearly four years, are you able to share (🤞🏻) any news around broader adoption of the proposed spec (or insight into if it may be accepted)?

PR #238, opened 2018-07-10
Document setting up extensionless feed paths

### LINKED ISSUE #237: Extensionless feed paths
For example, if you set the feed path to `feed`, `/feed` will return a `Content-Type` of `application/octet-stream` (instead of `text/xml`).

We could automatically place the feed in an `index.xml` (for example, `/feed/index.xml`) so the server sends the correct content type. Alternatively, we could document the `index.xml` workaround in the [Already have a feed path?](https://github.com/jekyll/jekyll-feed#already-have-a-feed-path) section.

See also #112 (more or less a duplicate).

sonicdoe:
Resolves #237.


README.md
```diff
@@ -45,6 +45,8 @@ feed:
 
 To note, you shouldn't have to do this unless you already have a feed you're using, and you can't or wish not to redirect existing subscribers.
 
+<span role="img" aria-label="Warning">⚠️</span> If your feed path does not end with `.xml`, you'll need to configure your web server to serve the proper content type. Alternatively, you can specify e.g. `feed/index.xml` to serve your feed from `/feed/`.
```
benbalter: I believe using `:warning:` here will result in similar markup.

sonicdoe: VoiceOver on macOS reads it slightly differently. However, I’ve noticed that GitHub converts it to their `<g-emoji>` custom element anyway, so I’ve changed it to `:warning:`. Thanks 👍

PR #240, opened 2018-09-09
Refactor to remove redundant calls and variables

ashmaroli:
- replace `hash[key] && hash[key][subkey]` with simpler Ruby 2.3's `hash.dig(key, subkey)`
- memoize `config` stashes
- prefer `hash.each_value` to avoid unnecessary allotment for unused block parameter
- prefer `Object#tap` over explicit local variable assignment (*cosmetic change*)


DirtyF:
T.I.L about [dig](https://www.tiagoamaro.com.br/2016/08/27/ruby-2-3-dig/) 👍

DirtyF:
@jekyllbot: merge +dev

PR #253, opened 2018-11-14
Allow customising collection feed titles

goulvench:
Allows customizing the feed title for a collection, in case you don't want to use the capitalized collection name.

### DIFF
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -12,9 +12,10 @@ module JekyllFeed
         Jekyll.logger.info "Jekyll Feed:", "Generating feed for #{name}"
         (meta["categories"] + [nil]).each do |category|
           path = feed_path(:collection => name, :category => category)
+          feed_title = feed_title(:collection => name, :category => category)
           next if file_exists?(path)
 
-          @site.pages << make_page(path, :collection => name, :category => category)
+          @site.pages << make_page(path, feed_title, :collection => name, :category => category)
         end
       end
     end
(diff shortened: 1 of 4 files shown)


goulvench:
If this is merged, it will allow generating links with the correct title for collection feeds (#254)

lib/jekyll-feed/generator.rb
```diff
@@ -48,6 +49,22 @@ def feed_path(collection: "posts", category: nil)
...
+    #
+    # collection - the name of a collection, e.g., "posts"
+    # category - a category within that collection, e.g., "news"
+    #
+    # Will return the site title or name
+    # ...followed by collection title or capitalized name
+    # ...followed by capitalized category name
+    def feed_title(collection: "posts", category: nil)
+      words = []
+      words << (@site.config["title"] || @site.config["name"])
```
ashmaroli:
```suggestion
      words << @site.config["title"] || @site.config["name"]
```

lib/jekyll-feed/generator.rb
```diff
@@ -48,6 +49,22 @@ def feed_path(collection: "posts", category: nil)
...
+      words << collections.dig(collection, "title") || collection.capitalize unless collection == "posts" # rubocop:disable Metrics/LineLength
```
ashmaroli:
```suggestion
      unless collection == "posts"
        words << collections.dig(collection, "title") || collection.capitalize
      end
```

lib/jekyll-feed/generator.rb
goulvench: That's what I had initially written, but due to operator precedence Ruby reads it as: "add title to words, or return name" and test `JekyllFeed parsing with site.name set uses site.name for the title` fails. If there's a more elegant syntax let me know.

lib/jekyll-feed/generator.rb
goulvench: Indeed, this is more readable. But we need parenthesis like above because of operator precedence, i.e: ` words << (collections.dig(collection, "title") || collection.capitalize)`

ashmaroli: I see. I've updated my suggestion. You can either commit that or push a local commit.
Entirely your call.

lib/jekyll-feed/generator.rb
ashmaroli: The more elegant way would be using a local variable. But that's just unnecessary allocation. So let's leave this as it is..
```ruby
title = @site.config["title"] || @site.config["name"]
words << title
```

PR #254, opened 2018-11-14
Output feed links for collections

goulvench:
Makes `{% feed_meta %}` generate a link for every collection.

All meta links have the same title at the moment, which is not good. But if my previous PR is accepted, it's just a matter of replacing `:title => title,` with `:title => generator.feed_title(:collection => collection, :category => category)` (and removing the `title` method from `meta-tag.rb`).

If there's a better way of grabbing a reference to the generator from meta-tag, I'd love to hear about it.


goulvench:
I'm open to discussion regarding this PR because it is a breaking change (an improvement in most cases, but still). Should `feed_meta` use parameters to make the new behavior opt-in?
Proposed syntax:
```
# Passing the collection name, requires one line per collection
{% feed_meta "updates" %}
# Boolean flag, if included all collections are listed
{% feed_meta including_collections %}
# Allow selecting which collections are included
{% feed_meta including_collections: "updates photos" %}
```
Feel free to comment, I'd be happy to implement it and update the PR.

goulvench:
Hi there, is there anything blocking this PR from being merged? I'll gladly update it if needed, just let me know what needs to be improved. Thanks!

lib/jekyll-feed/meta-tag.rb
```diff
@@ -17,21 +33,50 @@ def config
       @config ||= @context.registers[:site].config
     end
 
-    def attributes
+    def generator
+      @generator ||= @context.registers[:site].generators.select { |it| it.is_a? JekyllFeed::Generator }.first # rubocop:disable Metrics/LineLength
```
ashmaroli: Couple of things to change here:
  - RuboCop disabling comments should be placed above the method definition. A disabled cop should be enabled likewise, with the comment placed right after the filtered method(s).
  - Prefer using `Array#find` instead of chaining `Array#select#first`.
  - Block parameters should be immediately comprehensible. Even though it makes very little difference, a developer reading this code will spend a few seconds thinking if `it` means `"it"` or a shorthand for `"item"`.. :wink:
I'd prefer `generator` itself or `gen` for short.

```ruby
    def generator
      @generator ||= @context.registers[:site].generators.find do |generator|
        generator.is_a? JekyllFeed::Generator
      end
    end
```
```ruby
    def generator
      @generator ||= @context.registers[:site].generators.find { |gen| gen.is_a? JekyllFeed::Generator }
    end
```

lib/jekyll-feed/meta-tag.rb
```diff
@@ -17,21 +33,50 @@ def config
...
+    end
+
+    def link(collection, category)
+      attrs = attributes(collection, category).map { |k, v| %(#{k}="#{v}") }.join(" ")
```
ashmaroli: Since values from `attributes()` are always strings,
```suggestion
      attrs = attributes(collection, category).map { |k, v| "#{k}=#{v.inspect}" }.join(" ")
```

lib/jekyll-feed/meta-tag.rb
```diff
@@ -5,10 +5,26 @@ class MetaTag < Liquid::Tag
...
+    end
+
     def render(context)
       @context = context
-      attrs    = attributes.map { |k, v| %(#{k}="#{v}") }.join(" ")
-      "<link #{attrs} />"
+      if @args.strip == "include: all"
+        links = []
+        generator.collections.each do |collection, meta|
+          (meta["categories"] + [nil]).each do |category|
```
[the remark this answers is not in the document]
ashmaroli: This has shown to be a benign change via #262.

lib/jekyll-feed/meta-tag.rb
goulvench: Skipping this as I saw that this change was declined in #263.

lib/jekyll-feed/meta-tag.rb
```diff
@@ -5,10 +5,29 @@ class MetaTag < Liquid::Tag
...
+    end
+
     def render(context)
       @context = context
-      attrs    = attributes.map { |k, v| %(#{k}="#{v}") }.join(" ")
-      "<link #{attrs} />"
+      @collection = nil
+      @category   = nil
+
+      if @args.strip == "include: all"
```
ashmaroli: There's no need to `strip` again. It's been handled in the `initialize` method.
```suggestion
      if @args == "include: all"
```

ashmaroli:
@goulvench Thank you very much for addressing all of my concerns.
Just one final hurdle:

*I recently learned that YAML parses `yes || YES` as boolean `true` and `no || NO` as boolean `false`.*
How do we handle the following unlikely but probable configuration:
```yaml
feed:
  collections:
    collection:
      categories:
        - no
        - false
```

DirtyF:
@ashmaroli That's out of the scope of this PR, we could treat this elsewhere, as this could happen elsewhere in the config file

ashmaroli (dismissed):
LGTM! :+1:

ashmaroli:
This needs to have conflicts resolved..

goulvench:
Hi, I'd love to see this land in core but it's been more than 20 months since I opened this PR and I don't have time to work on it at the moment. Feel free to implement this in some way, fix the conflicts yourselves, or simply close this PR.

PR #263, opened 2019-02-04
Simplify interpolation by inspecting strings

ashmaroli:
- Since `%(#{k}="#{v}")` is a bit confusing to parse for a human..
- The other alternative is to escape double-quotes `"#{k}=\"#{v}\""` which I feel is uglier.
- `String#inspect` returns itself quoted and with any special characters escaped.


ashmaroli:
Okay then. I'll close this.

PR #264, opened 2019-02-17
Feed by tag

pmb00cs:
Hi I wanted to get jekyll to automatically generate feeds for the tags I apply to my blog. I've put together this pull request, which works functionally, but doesn't build with the current rubocop settings.
I'd like some help improving the work that I have done to make it acceptable to include in this project.

I'm not primarily a ruby developer, but a Sysadmin, so pointers would be welcomed.

Strictly my use case does not need this complexity, but it annoys me that Jekyll doesn't include this functionality that is part of most popular blogging platforms.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ The plugin will automatically use any of the following configuration variables, 
 
 * `title` or `name` - The title of the site, e.g., "My awesome site"
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
-* `url` - The URL to your site, e.g., `http://example.com`. If none is provided, the plugin will try to use `site.github.url`.
+* `url` - The URL to your site, e.g., `https://example.com`. If none is provided, the plugin will try to use `site.github.url`.
 * `author` - Global author information (see below)
 
 ### Already have a feed path?
@@ -181,6 +181,20 @@ feed:
         - news
         - updates
 ```
+
+## Excerpt Only flag
+
+Optional flag `excerpt_only` allows you to exclude post content from the Atom feed. Default value is `false` for backward compatibility.
+
+When in `config.yml` is `true` than all posts in feed will be without `<content>` tags.
+
+```yml
+feed:
+  excerpt_only: true
+```
+
+The same flag can be used directly in post file. It will be disable `<content>` tag for selected post.
+Settings in post file has higher priority than in config file.
 
 ## Tags
 
--- a/README.md
+++ b/README.md
@@ -210,7 +210,7 @@ If there are tags you don't want included in this auto generation you can exclud
 ```yml
 feed:
   tags:
-    excludes:
+    except:
       - tag-to-exclude
       - another-tag
 ```
@@ -228,7 +228,7 @@ If you only want to generate feeds for a few tags you can also set this.
 ```yml
 feed:
   tags:
-    includes:
+    only:
       - tag-to-include
       - another-tag
 ```
--- a/README.md
+++ b/README.md
@@ -151,6 +151,15 @@ feed:
     - updates
 ```
 
+## Posts limit
+
+By default the plugin limits the number of posts in the feed to 10. Simply define a new limit in your config:
+
+```yml
+feed:
+  posts_limit: 20
+```
+
 ## Collections
 
 Jekyll Feed can generate feeds for collections other than the Posts collection. This works best for chronological collections (e.g., collections with dates in the filenames). Simply define which collections you'd like feeds for in your config:
--- a/README.md
+++ b/README.md
@@ -182,6 +182,45 @@ feed:
         - updates
 ```
 
+## Tags
+
+To automatically generate feeds for each tag you apply to your posts you can add a tags setting to your config:
+
+```yml
+feed:
+  tags:
+```
+
+If there are tags you don't want included in this auto generation you can exclude them
+
+```yml
+feed:
+  tags:
+    excludes:
+      - tag-to-exclude
+      - another-tag
+```
+
+If you wish to change the location of these auto generated feeds (`/feed/by_tag/<TAG>.xml` by default) you can provide an alternative folder for them to live in.
+
+```yml
+feed:
+  tags:
+    path: "alternative/path/for/tags/feeds/"
+```
+
+If you only want to generate feeds for a few tags you can also set this.
+
+```yml
+feed:
+  tags:
+    includes:
+      - tag-to-include
+      - another-tag
+```
+
+Note that if you include a tag that is excluded a feed will not be generated for it.
+
 ## Contributing
 
 1. Fork it (https://github.com/jekyll/jekyll-feed/fork)
--- a/README.md
+++ b/README.md
@@ -188,7 +188,7 @@ To automatically generate feeds for each tag you apply to your posts you can add
 
 ```yml
 feed:
-  tags:
+  tags: true
 ```
 
 If there are tags you don't want included in this auto generation you can exclude them
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -40,7 +40,7 @@
   {% endif %}
 
   {% if page.tag %}
-    {% assign entries = site.tags.[page.tag] %}
+    {% assign entries = site.tags[page.tag] %}
   {% else %}
     {% assign entries = site[page.collection] %}
   {% endif %}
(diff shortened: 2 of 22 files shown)


pmb00cs:
So I've written some tests, there is still some work to do on making the code pass rubocop, and it would be nice to get some help with that

pmb00cs:
So I've done some work to refactor the logic I've built to reduce the rubocop errors, but the last error I'm not sure how to correct it without breaking things badly.

pmb00cs:
So I've fixed the last of the rubocop errors, and added extra tests for when the feature I'm wanting to add is not invoked (it occured to me that I hadn't added that test and I should have) to ensure it doesn't create feeds it shouldn't do. I would still appreciate any feedback you may have.

README.md
```diff
@@ -182,6 +182,45 @@ feed:
...
 ```
 
+## Tags
+
+To automatically generate feeds for each tag you apply to your posts you can add a tags setting to your config:
+
+```yml
+feed:
+  tags:
+```
```
ashmaroli: This is incorrect. This does not mean `{"feed"=>{"tags" => {}}}`
This evaluates to `{"feed"=>{"tags" => nil}}`

Instead, recommend enabling via a boolean `true`:
```yml
feed:
  tags: true
```

spec/jekyll-feed_spec.rb
```diff
@@ -470,4 +470,113 @@
...
+        expect(Pathname.new(dest_dir("feed/by_tag/success.xml"))).to_not exist
+      end
+    end
+
+
+    context "set tags setting" do
+      let(:overrides) do
+        {
+          "feed" => {
+            "tags" => {}
```
ashmaroli: AFAIK, it is not possible to obtain an empty Hash via YAML configuration.
So it'd be better if we test for explicit enabling via boolean `true`:
```suggestion
            "tags" => true
```

ashmaroli:
@pmb00cs I've refactored the Ruby code for you since it is not a familiar language to you and therefore would have simply led to noise if I were to just leave comments behind.

That said, I've marked two areas (documentation and tests) needing changes. Please make the changes when possible.
Feel free to further improve the code if you feel the need.

pmb00cs:
@ashmaroli thank you for your feedback, and for your changes. Your logic looks much cleaner than mine. I've fixed the comments you have made against the tests and the README

pmb00cs:
It's been a while since I raised this pull request, and it has been waiting on a review for most of that time. I could understand if there is reluctance to accept a new feature into your project, but I would like the opportunity to try and address any such reluctance if I can. As such is there anything I can do or say, to increase the chances of getting such a review? I apologise if the wait is simply down to work load however. I do know that open source project maintainers can be short on time.

lib/jekyll-feed/feed.xml
```diff
@@ -39,7 +39,12 @@
     </author>
   {% endif %}
 
-  {% assign posts = site[page.collection] | where_exp: "post", "post.draft != true" | sort: "date" | reverse %}
+  {% if page.tag %}
```
DirtyF: What about `tags`?

pmb00cs: You're right, this is inconsistent. I use "tags" just about everywhere else. Let add a commit to make this more consistent.

ashmaroli:
@pmb00cs Can you add a test-case for a scenario involving front matter key `tag`? Since Jekyll handles `tag` and `tags` automatically, there shouldn't be an issue. But it helps to be sure.
```yaml
---
tag: foobar   # value is a string instead of an array..
---

This a test document.
```

pmb00cs:
That looks like a reasonable ask. I'll see what I can do.

pmb00cs:
I've added the requested test. It's passing for me locally.

ashmaroli:
Thank you very much @pmb00cs.

@DirtyF This LGTM! as well. I'm not leaving an explicit approval here since I've made some non-trivial changes. If none of the other maintainers requests for further changes in a couple of days, feel free to merge.

DirtyF:
@ashmaroli yes, I would prefer a former approval from @mattr- @parkr or @benbalter as this plugin is white listed on GitHub Pages.

lib/jekyll-feed/generator.rb
```diff
@@ -69,6 +70,31 @@ def collections
       @collections
     end
 
+    def generate_feed_by_tag
+      tags_config = config["tags"].is_a?(Hash) ? config["tags"] : {}
+
+      excludes  = tags_config["excludes"] || []
+      includes  = tags_config["includes"] || @site.tags.keys
+      tags_pool = includes - excludes
+      tags_path = tags_config["path"] || "/feed/by_tag/"
```
benbalter: What happens if I pass in a malicious path here via the site's config? (e.g., `../../../../../some-secret-file`)?

* I _think_ `file_exists?` below could reveal the file's existence?
* It [looks like](https://github.com/jekyll/jekyll/blob/master/lib/jekyll/page.rb#L46) we may be sanitizing the path to be _within the plugin source_? Is that possible or am I misreading?
* If that's the case, then a malicious user may be able to write to an arbitrary path (potentially within the plugin source?)

pmb00cs: It gets passed to the same "make_page" function as generating the main feed uses. Honestly I'm not sure what happens with malicious paths get used, I assumed that function already included sanity checking for the path. The only change I made to that function was to allow it to include the option to pass in the tag you want a feed for.

ashmaroli: @benbalter In reality the `@base` and `@path` values never get used:
```ruby
# jekyll/lib/jekyll/page.rb

def path
  data.fetch("path") { relative_path }
end

# The path to the page source file, relative to the site source
def relative_path
  @relative_path ||= File.join(*[@dir, @name].map(&:to_s).reject(&:empty?)).sub(%r!\A\/!, "")
end
```
`@base` which equals to the value of `__dir__` is not used in our use-case because the `:read_yaml` is overridden:
https://github.com/jekyll/jekyll-feed/blob/9cda8c4e4940cc6947f1525b23461d06b1caaa1c/lib/jekyll-feed/page-without-a-file.rb#L5-L7

That said, your observation cannot be left disregarded (also present in the `master` and published versions) since the instance variable could still be read via another third-party plugin when used outside the GitHub Pages environment.

ashmaroli: > What happens if I pass in a malicious path here via the site's config? (e.g., ../../../../../some-secret-file)?

Added a test for this in https://github.com/jekyll/jekyll-feed/pull/264/commits/93c69a83a3fc2a8bd80d95f5348cdb5c8eecdd7a

ashmaroli:
@pmb00cs I was wondering.. How about renaming `config["includes"]` to `config["only"]` so that it is more intuitive..? `includes` gives the feeling that it is to *include* a tag that is otherwise not considered... and then `config["excludes"]` to `config["except"]`..?

pmb00cs:
That looks like a sensible suggestion.

pmb00cs:
I've pushed up a change to address that @ashmaroli suggested to make the settings clearer on what they do.

FREELANCEGITIT:
@benbalter @DirtyF @pmb00cs @ashmaroli @mojombo 
Nice work. I enjoyed reading your correspondence and am learning a lot c
From this process. Best, K

pmb00cs:
Is there any update on when the outstanding review of this pull request might happen?

hozza:
I would super love this to be merged! Perfect for connecting a Jekyll site to MailChimp for tag based email marketing! 🤩

spec/jekyll-feed_spec.rb
```diff
@@ -497,6 +497,160 @@
...
+          expect(Pathname.new(dest_dir("feed/by_tag/success.xml"))).to_not exist
+
+          expect(Pathname.new(dest_dir("questionable/path/test.xml"))).to exist
+          expect(Pathname.new(dest_dir("questionable/path/fail.xml"))).to exist
+          expect(Pathname.new(dest_dir("questionable/path/success.xml"))).to exist
+        end
+      end
+    end
+
+    context "test nonarray tag value" do
```
parkr, line 635: If I understand the context of this test properly, this is to test Jekyll Core functionality – namely, that `tag: value` becomes `[“value”]`. I don’t think we need this test since the Jekyll Core test suite should cover this, right?

ashmaroli:
@pmb00cs Will you be able to update the tests in light of Parker's review?

pmb00cs:
Hi, @ashmaroli I've removed the test that @parkr has raised as unneeded, and also merged the latest commits in master to ensure this pull request is fully up to date.

DirtyF:
@jekyll: merge +minor 🎉

PR #273, opened 2019-03-29
 Add media:content tag to support Mailchimp

mikepsinn:
To help ensure that MailChimp can view your feed’s images, you can include medium or type information inside the `<media:content>` tag.

Example: `<media:content url=”http://hathaway.edu/example.jpg” medium="image">`

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -86,7 +86,7 @@
           {% assign post_image = post_image | absolute_url %}
         {% endunless %}
         <media:thumbnail xmlns:media="http://search.yahoo.com/mrss/" url="{{ post_image | xml_escape }}" />
-        <media:content url="{{ post_image | xml_escape }}" medium=”image”>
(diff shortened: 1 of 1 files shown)


lib/jekyll-feed/feed.xml
```diff
@@ -86,6 +86,7 @@
           {% assign post_image = post_image | absolute_url %}
         {% endunless %}
         <media:thumbnail xmlns:media="http://search.yahoo.com/mrss/" url="{{ post_image | xml_escape }}" />
+        <media:content url="{{ post_image | xml_escape }}" medium="image">
```
mattr-: Looks like you're missing a proper closing tag here, which we can get by just adding a `/` before the `>`

```suggestion
        <media:content url="{{ post_image | xml_escape }}" medium="image" />
```

mikepsinn: Thanks, @mattr-!  Sorry, I just saw this but it looks like someone else already took care of it.

PR #276, opened 2019-06-09
feeds for all

arghc:
- Adjusted feed.xml to recognize category and categories.
- Added ability to generate feeds for tags.
- Added flag to generate feeds for all categories and/or tags.


DirtyF:
We just merged #153

Issue #277, opened 2019-07-03
Excerpt Only feed


davidbstein:
Many of my posts have interactive components that don't inline properly in the atom feed. Is there a way to introduce an "excerpt only" flag for posts that don't interact well with RSS readers?

jekyllbot:
This issue has been automatically marked as stale because it has not been commented on for at least two months.

The resources of the Jekyll team are limited, and so we are asking for your help.

If this is a **bug** and you can still reproduce this error on the <code>master</code> branch, please reply with all of the information you have about it in order to keep the issue open.

If this is a feature request, please consider whether it can be accomplished in another way. If it cannot, please elaborate on why it is core to this project and why you feel more than 80% of users would find this beneficial.

This issue will automatically be closed in two months if no further activity occurs. Thank you for all your contributions.

psfrolov:
I would also like to see an ability to use excerpt-only feeds. Rendering whole article in atom is actually suitable only for very simple content. For example, I have the following problems with full articles in atom produced by jekyll-feed: 
* SVG diagrams are not rendered (broken image icon appears instead)
* Most of styling (syntax hihlightning, typography features) are lost.
* User is unable to comment on the article (no JavaScript).

torrocus:
I prepared PR to this feature. It's very basic, but worst is better.
https://github.com/jekyll/jekyll-feed/pull/287

torrocus:
Hey! Is any of you (@pathawks & @DirtyF) still responsible for this project?

ffes:
With #287 merged this issue can be closed, right?

ashmaroli:
Yes, it can be closed. Thanks for letting us know @ffes

PR #287, opened 2019-10-07
Excerpt only flag

torrocus:
Create new flag `excerpt_only`, which allows to exclude post content from feed.

PR to issue: https://github.com/jekyll/jekyll-feed/issues/277

BTW I don't like name `excerpt_only`, but don't have better idea. IMO this name should get true value for full content and false value for excerpt.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,21 @@ feed:
         - updates
 ```
 
+## Excerpt Only flag
+
+Optional flag `excerpt_only` allows you to exclude post content from the Atom feed. Default value is `false` for backward compatibility.
+
+When in `config.yml` is `true` than all posts in feed will be without `<content>` tags.
+
+```yml
+feed:
+  excerpt_only: true
+```
+
+The same flag can be used directly in post file. It will be disable `<content>` tag for selected post.
+Settings in post file has higher priority than in config file.
+
+
 ## Contributing
 
 1. Fork it (https://github.com/jekyll/jekyll-feed/fork)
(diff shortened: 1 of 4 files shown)


ashmaroli:
@torrocus First of all, apologies for not reviewing this earlier.
While I don't have an issue with the idea, I'm on the fence regarding the implementation.

Since Parker has already approved this, I won't block this from being merged but would love it if you could clarify the following / make some changes:
- `| default: false }}` is unnecessary. The subsequent conditional is valid even if `excerpt_only` evaluates to `nil`.
- Why conditionally `assign post_content` and then immediately test `post_content` for a truthy? Why not render directly?
  ```xml
  {% unless excerpt_only %}
    <content type="html" xml:base="{{ post.url | absolute_url | xml_escape }}">
      {{ post.content | strip | xml_escape }}
    </content>
  {% endunless %}
  ```

torrocus:
@parkr Thanks for dig.
@ashmaroli No problem, it's good to know someone is coming here.

Regarding your attention with `| default: false }}` you're right. I deleted the redundant default value. I wanted to show what the default value is. But later I included this information in the documentation and it should be enough.

Your second point is not that simple. Post does not always have content. The question is is this correct? View to file spec/fixtures/_posts/2015-02-12-strip-newlines.md
Personally, I would prefer render directly. But I do not know the whole context and did not want to interfere with existing functionalities.

ashmaroli:
> Post does not always have content

To my knowledge, the `content` attribute is never `nil` or anything other than a String. It can be just empty `""` but in Ruby, that is considered a truthy as well.

Of course, having a test for this would make things clear and concrete.

torrocus:
Right, I corrected it. Previously, one test failed. Now everything is OK, and I already had this code written in this way.

torrocus:
Please squash all commits when you will merge.

ashmaroli:
@parkr Since the diff changed significantly after you gave your approval, could you do a *re-review*?
Thanks.

DirtyF:
@jekyllbot: merge +minor

PR #297, opened 2020-01-22
Add support for page.description in front matter to become entry `<summary>`

### LINKED ISSUE #216: Use page.description for summary element, falling back to page.excerpt
[Jekyll SEO Tag](https://github.com/jekyll/jekyll-seo-tag) prioritises the `page.description` front matter value over `page.excerpt` ([see PR](https://github.com/jekyll/jekyll-seo-tag/commit/6dab82639ea393f93033efc5e3f5005e864de2b4)). It would be great to also have this for the feed item's `summary` element.

While `excerpt` typically contains a truncated snippet from the beginning of the post, `description` provides a place for the author to better summarise the content.

mohd-akram:
Closes #216.

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,6 +62,8 @@ Additionally, the plugin will use the following values, if present in a post's Y
 
 * `author` - The author of the post, e.g., "Dr. Jekyll". If none is given, feed readers will look to the feed author as defined in `_config.yml`. Like the feed author, this can also be an object or a reference to an author in `_data/authors.yml` (see below).
 
+* `description` - A short description of the post.
+
 ### Author information
 
 *TL;DR: In most cases, put `author: [your name]` in the document's front matter, for sites with multiple authors. If you need something more complicated, read on.*
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -94,8 +94,9 @@
         <category term="{{ tag | xml_escape }}" />
       {% endfor %}
 
-      {% if post.excerpt and post.excerpt != empty %}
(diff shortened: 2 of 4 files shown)


parkr:
I’m not sure I understand why you would use a different field here, namely why you would use `description` instead of `excerpt`. Jekyll allows you to overwrite `excerpt` in front matter, so why not just overwrite the excerpt for the post or page in question?

mohd-akram:
I use `description` because it's supported by `jekyll-seo-tag` (see the issue above) and because overriding `excerpt` feels weird. `title` and `description` together are also analogous to their HTML counterparts.

benbalter:
👋😄 I came here to request the same feature. 

Since Jekyll SEO Tag supports `title` + `description`, I add short 1-2 sentence descriptions to each post which serve as a concise summary of the post's content. By using `excerpt` and only `excerpt`, Jekyll Feed either only outputs the post body up to the first line break or requires site maintainers to duplicate summaries across both the `excerpt` and `description` fields for all posts. 

:+1: to preferring `description` if it's present to bring things a bit closer in line with the way Jekyll SEO Tag treats front matter.

parkr:
jekyll-seo-tag has a field called "description" which uses `page.description` then `page.excerpt` and finally falls back  to the site description:

https://github.com/jekyll/jekyll-seo-tag/blob/c658a8dd9d94520c090547b164e643441df8e62f/lib/jekyll-seo-tag/drop.rb#L90-L94

parkr (approved):

parkr:
@jekyllbot: merge +minor

PR #300, opened 2020-02-17
Fix issue 299

pingu8007:
Entry.Author element will be filled with:
- name
- twitter handle, prepend with `@`
- key string used to search in authors.yml, if neither name nor twitter available

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -59,7 +59,11 @@
       {% assign post_author = site.data.authors[post_author] | default: post_author %}
       {% assign post_author_email = post_author.email | default: nil %}
       {% assign post_author_uri = post_author.uri | default: nil %}
-      {% assign post_author_name = post_author.name | default: post_author %}
+      {% assign post_author_twitter = post_author.twitter | default: nil %}
+      {% if post_author_twitter %}
+        {% assign post_author_twitter = post_author.twitter | prepend: "@" %}
+      {% endif %}
+      {% assign post_author_name = post_author.name | default: post_author_twitter | default: post_author %}
 
       <author>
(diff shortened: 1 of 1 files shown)


ashmaroli:
Hi @pingu8007 Could you update the title to something that best describes the patch? Thanks.
*(One should not have to read an entire issue-ticket to get that info..)*

lib/jekyll-feed/feed.xml
```diff
@@ -56,10 +56,18 @@
...
+        {% assign post_author_name = post_author %}
+      {% endunless %}
       {% assign post_author = site.data.authors[post_author] | default: post_author %}
       {% assign post_author_email = post_author.email | default: nil %}
       {% assign post_author_uri = post_author.uri | default: nil %}
-      {% assign post_author_name = post_author.name | default: post_author %}
+      {% assign post_author_twitter = post_author.twitter | default: nil %}
+      {% if post_author_twitter %}
+        {% assign post_author_twitter = post_author.twitter | prepend: "@" %}
+      {% endif %}
```
ashmaroli: I don't understand why this logic is so complex.
Isn't the following equivalent (to ln66 - ln69):
```
{% if post_author.twitter %}
  {% assign post_author_twitter = post_author.twitter | prepend: "@" %}
{% else %}
  {% assign post_author_twitter = nil %}
{% endif %}
```

pingu8007: @ashmaroli You are right, I didn't thought about it.
The outer if block was added after I figure out the prepend filter break the nil check. I will update it later.

ashmaroli: Thanks. While you update it, please update the logic for `post_author_name` as well.

PR #305, opened 2020-04-08
Fix feed link when post title contains HTML

guilhermesimoes:
This patch fixes the following warning:

> This feed is valid, but interoperability with the widest range of feed readers could be improved by implementing the following recommendations:
> - link should not contain HTML

as per https://validator.w3.org/feed/

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -45,8 +45,10 @@
   {% endif %}
   {% for post in posts limit: 10 %}
     <entry{% if post.lang %}{{" "}}xml:lang="{{ post.lang }}"{% endif %}>
-      <title type="html">{{ post.title | smartify | strip_html | normalize_whitespace | xml_escape }}</title>
-      <link href="{{ post.url | absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
+      {% assign post_title = post.title | smartify | strip_html | normalize_whitespace | xml_escape %}
+
+      <title type="html">{{ post_title }}</title>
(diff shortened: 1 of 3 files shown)


lib/jekyll-feed/feed.xml
```diff
@@ -45,8 +45,10 @@
   {% endif %}
   {% for post in posts limit: 10 %}
     <entry{% if post.lang %}{{" "}}xml:lang="{{ post.lang }}"{% endif %}>
-      <title type="html">{{ post.title | smartify | strip_html | normalize_whitespace | xml_escape }}</title>
-      <link href="{{ post.url | absolute_url }}" rel="alternate" type="text/html" title="{{ post.title | xml_escape }}" />
+      {% assign post_title = post.title | smartify | strip_html | normalize_whitespace | xml_escape %}
+
+      <title type="html">{{ post_title }}</title>
+      <link href="{{ post.url | absolute_url }}" rel="alternate" type="text/html" title="{{ post_title }}" />
```
ashmaroli, line 51: I can see that the `href` attribute hasn't changed from the existing value on `master`.
But shouldn't that have an ` | xml_escape` as well..?

/cc @pathawks

guilhermesimoes: Would you like me to fix that as well? I can do that

ashmaroli: I know *you can*.., was expecting a confirmation from @pathawks though..

DirtyF:
@jekyll: merge +fix

lib/jekyll-feed/feed.xml
DirtyF: Feel free to fix this in another issue. thanks.

PR #306, opened 2020-04-11
XML escape the title field of feed_meta

Lax:
When site title has special characters such as `<&>"`, this patch would encode them with xml compatible format(`&lt;&amp;&gt;`)


mattr-:
@jekyllbot: merge +minor

PR #307, opened 2020-04-13
chore(deps): require Ruby >=2.4.0

DirtyF:
Align with other core plugins that already require 2.4 
Ruby 2.4 is now EOL.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -12,7 +12,6 @@ environment:
       JEKYLL_VERSION : ">= 4.0.0.pre.alpha1"
     - RUBY_FOLDER_VER: "26"
     - RUBY_FOLDER_VER: "24"
-    - RUBY_FOLDER_VER: "23"
 
 install:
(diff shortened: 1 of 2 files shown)


ashmaroli:
From what I recall, we have not started enforcing Ruby 2.4 in core plugin's gemspec if the library code doesn't use any Ruby 2.4+ specific syntax..

DirtyF:
We don't use Ruby 2.4 syntax unless we can :)

DirtyF:
https://www.ruby-lang.org/en/news/2020/04/05/support-of-ruby-2-4-has-ended/

ashmaroli:
I'm still of the opinion that this change is not necessary &mdash; there's *no explicit need* to use this plugin with Ruby 2.4 and above. But there's already an approval, so perhaps I'm not seeing the value..

DirtyF:
@jekyll: merge +minor

PR #309, opened 2020-04-27
chore: use Dir to list source files

DirtyF:
https://piotrmurach.com/articles/writing-a-ruby-gem-specification/

### DIFF
--- a/jekyll-feed.gemspec
+++ b/jekyll-feed.gemspec
@@ -3,17 +3,18 @@
 require_relative "lib/jekyll-feed/version"
 
 Gem::Specification.new do |spec|
-  spec.name          = "jekyll-feed"
-  spec.version       = Jekyll::Feed::VERSION
-  spec.authors       = ["Ben Balter"]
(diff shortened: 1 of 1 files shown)


ashmaroli:
Is your motivation behind this to follow a convention or to just bundle the necessary files in the gem?
If it is the latter case, then IMHO the gem contents ought to roughly be:
```diff
  lib/**/*
- script/**/*
- spec/**/*
- .gitignore
- .rspec
- .rubocop.yml
- .travis.yml
- appveyor.yml
- Gemfile
  History.markdown
- Rakefile
  README.md
```

DirtyF:
@ashmaroli looks like it would be easier to only add lib and Markdown files :)

ashmaroli:
Yep. `:select` instead of `:reject`.. :wink:

DirtyF:
@ashmaroli found an even simpler way to list files with `Dir` thanks to @piotrmurach

ashmaroli:
@jekyllbot: merge +dev

PR #316, opened 2020-07-10
Add support for drafts

torrocus:
From now, when you run `jekyll serve --drafts`, your RSS feeds will also contain also drafts posts.
The same behavior is in Jekyll Sitemap.

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -40,14 +40,17 @@
   {% endif %}
 
   {% if page.tags %}
-    {% assign entries = site.tags[page.tags] %}
+    {% assign posts = site.tags[page.tags] %}
   {% else %}
-    {% assign entries = site[page.collection] %}
+    {% assign posts = site[page.collection] %}
   {% endif %}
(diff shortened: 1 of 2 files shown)


DirtyF:
What's the point of supporting drafts for a local feed?

torrocus:
I think this is useful in the deployment process. It's more convenient to create drafts than real posts to test different things.
When I enable --drafts options, I expect:
- in the index I will see posts from the _drafts and _posts folders
- sitemap will contain links to all these posts
- feed will contain the latest (according to the limit) posts including _drafts folder

At present, this behavior is inconsistent. Index is ok, sitemap is ok.
RSS feed behaves non-intuitively.

torrocus:
That's what I meant. I wanted to use what Jekyll already has.
And I was surprised that it doesn't work as expected.

DirtyF:
@jekyll: merge +minor

DirtyF:
@torrocus released in [0.15.0](https://github.com/jekyll/jekyll-feed/releases/tag/v0.15.0 :shipit:

PR #318, opened 2020-08-02
The RSS feed breaks on special chars

playest:
The RSS feeds breaks if a "&" is in the file name of the post. For example a file named "2020-08-01-Hello so&so.md" would have broken the feed without this commit.


ashmaroli:
Hi, could you add a test for this change?
Thanks.

playest:
I'll try. I don't really know how though. I just patched the file feed.xml on local version of the code.
I'll look into it.

PR #321, opened 2020-09-14
Support all tags

somini:
Instead of allowing only a subset of tag names, `slugify` the tag name
so that the URL is always valid.

This is the same thing done on "jekyll-archives":
- https://github.com/jekyll/jekyll-archives/blob/master/lib/jekyll-archives/archive.rb#L132

There is a possibility that the same tag slugifies to the same string, but it's probably fine.


somini:
This is already done on the latest release.

somini:
> This is already done on the latest release.

My mistake.

somini:
Rebased to the latest master.

somini:
Rebased to the latest master.

gravitystorm:
I agree that this is a better approach. It's certainly better than silently dropping tags!

I'm not a maintainer here, so I can't recommend what to do next. But my suggestion would be to add some tests (e.g. tags with spaces, tags with underscores, etc), since I wouldn't be comfortable myself merging a PR without them.

Finally, I think we would also need to consider backwards compatibility. At the moment, the feeds have case-sensitive paths, and the tags can contain underscores, and there will be people who have published their feeds like this already, and those feeds will need to keep working.

|tag|current|slugify|
|-|-|-|
|foo|foo.xml|foo.xml|
|BAR|BAR.xml|bar.xml|
|foo_bar|foo_bar.xml|foo-bar.xml|
|foo-baz| (no output)|foo-baz.xml|
|with space| (no output)|with-space.xml|

I dislike adding configuration options, but maybe it needs a backwards-compatibility flag?

somini:
I agree, a configuration

somini:
Rebased this to the latest master.

somini:
4fa351acf3be5ddcaccbeb7ddb1649465f51e546 extends this to collection and category names.

Issue #325, opened 2020-10-02
XML escaping the site title with jekyll-github-metadata errors


parkr:
The jekyll-github-metadata plugin can automatically add a `site.title`. It does this lazily so that it only fetches the data it needs when it's requested. One such request is in `JekyllFeed::MetaTag#render`, where it does `attributes.map`. It currently attempts to encode the value as XML:

https://github.com/jekyll/jekyll-feed/blob/c8e7a43d6fa4e34abf90ff1cea00d1b50df6dfb4/lib/jekyll-feed/meta-tag.rb#L10

When `jekyll-github-metadata` populates the `title` key, it adds a `Jekyll::GitHubMetadata::Value` object, which does **not** have an `#encode` method. This means that anyone using the jekyll-github-metadata plugin with Jekyll-feed is seeing their site not build properly. 

One simple solution here is to cast to a string with `#to_s` if the attribute doesn't respond to `#encode`.

/cc introduced in: https://github.com/jekyll/jekyll-feed/pull/306

PR #326, opened 2020-10-02
MetaTag: when encoding for XML special characters, handle non-string objects

parkr:
Fixes https://github.com/jekyll/jekyll-feed/issues/325.

The jekyll-github-metadata plugin, for example, sets `site.title` and
`site.name` to `Jekyll::GitHubMetadata::Value` objects which respond to
`#to_s` and `#to_liquid`, but NOT `#encode`. Therefore, we should cast
to a string if `#encode` is not yet available.

The reason we cast to a string is because the previous code was
`"#{k}=#{v}"`, which was implicitly casting to a string for `v`.

### DIFF
--- a/lib/jekyll-feed/meta-tag.rb
+++ b/lib/jekyll-feed/meta-tag.rb
@@ -7,8 +7,11 @@ module JekyllFeed
 
     def render(context)
       @context = context
-      attrs    = attributes.map { |k, v| %(#{k}=#{v.encode(:xml => :attr)}) }.join(" ")
-      "<link #{attrs} />"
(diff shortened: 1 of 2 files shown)


DirtyF (approved):
Crystal clear ✨

ashmaroli:
@parkr Do we really need to have that `respond_to?(:encode)` check?
Can't we cast and proceed right away?
```ruby
attributes.map { |k, v| "#{k}=#{v.to_s.encode(:xml => :attr)}" }.join(" ")
```

parkr:
No worries at all 😄 

@jekyllbot: merge +dev

PR #327, opened 2020-10-04
Fold private methods into the `:render` method as local variables

ashmaroli:
Since the meta tag contents doesn't depend on *page-specific* metadata, the resulting markup is essentially the same for all pages / documents in a site:
https://github.com/jekyll/jekyll-feed/blob/653999c975564e30003d2c0c247d59e5a89eecda/lib/jekyll-feed/meta-tag.rb#L23-L38

Therefore in theory, the result can be safely memoized to avoid generating arrays for every page / document..

### DIFF
--- a/lib/jekyll-feed/meta-tag.rb
+++ b/lib/jekyll-feed/meta-tag.rb
@@ -6,15 +6,18 @@ module JekyllFeed
     include Jekyll::Filters::URLFilters
 
     def render(context)
-      @context = context
-      attrs    = attributes.map do |k, v|
-        v = v.to_s unless v.respond_to?(:encode)
-        %(#{k}=#{v.encode(:xml => :attr)})
-      end
-      "<link #{attrs.join(" ")} />"
+      @context ||= context
+      memoized_result
     end
 
     private
+
+    def memoized_result
+      @memoized_result ||= begin
+        attrs = attributes.map { |k, v| "#{k}=#{v.to_s.encode(:xml => :attr)}" }
+        "<link #{attrs.join(" ")} />"
+      end
+    end
 
     def config
       @config ||= @context.registers[:site].config
--- a/lib/jekyll-feed/meta-tag.rb
+++ b/lib/jekyll-feed/meta-tag.rb
@@ -8,30 +8,20 @@ module JekyllFeed
     def render(context)
       # Jekyll::Filters::URLFilters requires `@context` to be set in the environment.
       @context = context
-      xml_encoded_link_tag
-    end
 
-    private
+      config = context.registers[:site].config
+      path   = config.dig("feed", "path") || "feed.xml"
+      title  = config["title"] || config["name"]
 
-    def config
-      @config ||= @context.registers[:site].config
-    end
+      attributes = {
+        :type => "application/atom+xml",
+        :rel  => "alternate",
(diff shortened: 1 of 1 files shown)


lib/jekyll-feed/meta-tag.rb
```diff
@@ -6,16 +6,19 @@ class MetaTag < Liquid::Tag
     include Jekyll::Filters::URLFilters
 
     def render(context)
-      @context = context
-      attrs    = attributes.map do |k, v|
-        v = v.to_s unless v.respond_to?(:encode)
-        %(#{k}=#{v.encode(:xml => :attr)})
-      end
-      "<link #{attrs.join(" ")} />"
+      @context ||= context
```
parkr, line 9: Since `context` could theoretically change (and therefore the code should reflect this), I'd feel more comfortable with an approach using the context's hash (the number Ruby assigns to each unique object):

```ruby
def render(context)
  link_tag(context)
end

def link_tag(context)
  @link_tags[context.hash] ||= "<link #{xml_encoded_attributes(context).join(" ")} />"
end
```

etc.

In this model, we always pass `context` down through the methods, making them pure functions, instead of relying on `@context` being set (and therefore acting more like Object methods relying on state). What do you think about an approach like this?

ashmaroli: From the source-code on `master`, the only place `@context` is ever used in this Liquid tag instance is to get the site config:
https://github.com/jekyll/jekyll-feed/blob/653999c975564e30003d2c0c247d59e5a89eecda/lib/jekyll-feed/meta-tag.rb#L19-L21
Since the site config shouldn't be changing once `site.render` phase starts, memoizing with a hash-store therefore seems wasteful to me...

ashmaroli: `@context` is required to be set for `Jekyll::Filters::URLFilters` to work.
So the suggested model is not applicable.

parkr: > Since the site config shouldn't be changing once `site.render` phase starts, memoizing with a hash-store therefore seems wasteful to me...

The code as it exists today has a `render` method that takes in a `context`. That means that `render` should respond to changes in `context`. From a strict read of just the method signatures, one would expect that inputting new information into `context` and passing it to `render` would yield a different result. I think the only way to fix this that can use an object-method approach is to do `MetaTag.new(context).render` and update everything accordingly. The problem stems from the programmer's expectation that `f(input)` should always take `input` into account rather than just the input to the first call to `f()`. This hybrid it's-a-function-but-we-treat-it-like-an-object relies on implicit understanding of the implementation, which could change. The function signatures should reflect this and move fully into an object approach, or use a pure function approach instead.

ashmaroli: Okay @parkr.

To avoid further bike-shedding, I've decided to not memoize at all..
(*since this tag is typically used once per `page` and hence the `context.hash` is never constant*)

I have refactored the whole implementation to not call private methods at all.

Is the current diff acceptable to you?

hugoglezcon (commented):
Gracias pero necesario ver el conte

parkr (approved):

parkr:
@jekyllbot: merge +bug

PR #332, opened 2021-01-17
Switched xml_escape for <![CDATA[]]> for post content

michaelnordmeyer:
See "Handling Atom Text and Content Constructs" (http://www.xml.com/pub/a/2005/12/07/handling-atom-text-and-content-constructs.html) for an explanation why this is valid and the recommended way of handling unescaped content in XML documents.

TL;DR: Much nicer to read if you for some reason have a look at the feed.xml source. Also less parsing on the client-side.


parkr (approved):
Needs some tests fixed, but 👍

michaelnordmeyer:
I found just one failed test and fixed it.

parkr (approved):

parkr:
@jekyllbot: merge +fix

PR #337, opened 2021-02-18
Add Ruby 3.0 to CI

oturpe:
jekyll-feed will be [included in Fedora 34][pkg] which will also use [Ruby 3.0
as system Ruby interpreter][ruby]. This combination is known to work, so it
would make sense to add it to CI.

[pkg]: https://src.fedoraproject.org/rpms/rubygem-jekyll-feed
[ruby]: https://fedoraproject.org/wiki/Changes/Ruby_3.0

### DIFF
--- a/jekyll-feed.gemspec
+++ b/jekyll-feed.gemspec
@@ -24,7 +24,6 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "nokogiri", "~> 1.6"
   spec.add_development_dependency "rake", "~> 12.0"
   spec.add_development_dependency "rspec", "~> 3.0"
-  spec.add_development_dependency "rss" if RUBY_VERSION >= '3.0.0'
   spec.add_development_dependency "rubocop-jekyll", "~> 0.5"
   spec.add_development_dependency "typhoeus", ">= 0.7", "< 2.0"
 end
--- a/jekyll-feed.gemspec
+++ b/jekyll-feed.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "nokogiri", "~> 1.6"
   spec.add_development_dependency "rake", "~> 12.0"
   spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency "rss" if RUBY_VERSION >= '3.0.0'
   spec.add_development_dependency "rubocop-jekyll", "~> 0.5"
   spec.add_development_dependency "typhoeus", ">= 0.7", "< 2.0"
 end
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,16 @@
 language: ruby
 cache: bundler
 rvm:
-  - &latest_ruby 2.7
+  - 3.0
+  - 2.7
   - 2.5
 env:
   matrix:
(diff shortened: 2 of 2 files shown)


oturpe:
Both Travis and AppVeyor are in use. The reason for this is unclear for me. I updated only Travis, because it had newer versions of everything. Should also AppVeyor be updated?

DirtyF (approved):

oturpe:
Oh, build for Ruby 3.0 failed due to missing `rss` gem. In Ruby 3.0, it [became a bundled gem][bundled-gem], thus it needs to be added as a dependency. I will fix that in this pull request. On the other hand, I see that there is some GitHub actions work going on the master branch, does that mean that Travis on its way out?

[bundled-gem]: https://www.ruby-lang.org/en/news/2020/12/25/ruby-3-0-0-released/

DirtyF:
@oturpe thanks I'm getting the same rss error on GitHub Actions, and yes the plan is to move all our CI to GitHub Actions

oturpe:
Rss gem fix added. Very common problem in Ruby 3.0 migration, because there was no warning in Ruby 2.7. Luckily, easy to fix too.

You can either merge this pull request, or just take the good parts if that fits the plan better.

jekyll-feed.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |spec|
   spec.add_development_dependency "nokogiri", "~> 1.6"
   spec.add_development_dependency "rake", "~> 12.0"
   spec.add_development_dependency "rspec", "~> 3.0"
+  spec.add_development_dependency "rss" if RUBY_VERSION >= '3.0.0'
```
ashmaroli, line 27: This change is not correct.
A gemspec is evaluated only during `gem build` calls or via a Gemfile pointing to the gem's repository.

In other words, if I were to cut the gem release using Ruby 2.7, then `rss` won't be marked as a dependency.

ashmaroli (changes requested):

jekyll-feed.gemspec
DirtyF: Isn't that the whole point since `rss`gem is only required since Ruby 3.0?

oturpe: I thought `rss` only needed for tests, in `spec/spec_helper.rb`. At least I could not find any other place where it is required. So the idea was: If on Ruby 3, add `rss` so tests can run. `if` can be removed, I was just unsure if there is some constraint preventing from things working on older Ruby versions.

ashmaroli: > I thought `rss` is only needed for tests, in `spec/spec_helper.rb`

Yes, that is correct.
To maintain consistency when cutting the release on differing Ruby versions, I recommend moving the `rss` gem declaration to the Gemfile.

DirtyF, line 27:
```suggestion
```

oturpe:
The solution for the `rss` problem that went is better than my proposal. Ruby 3.0 is now included in ci, so what I was trying to achieve with this pull request has been achieved. Thank you for your effort!

ashmaroli (approved):

ashmaroli:
Thank you @oturpe 
@jekyllbot: merge +dev

PR #338, opened 2021-03-09
Integrate Webfeeds for better Feedly integration

dylmye:
Integrate Webfeeds as per [this Feedly blogpost](https://blog.feedly.com/10-ways-to-optimize-your-feed-for-feedly/) and [Mark Mitchell's article](https://markonproduct.com/coding-an-atom-feed-optimized-for-feedly)

I've also updated the readme describing the values

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,6 +34,18 @@ The plugin will automatically use any of the following configuration variables, 
 * `url` - The URL to your site, e.g., `https://example.com`. If none is provided, the plugin will try to use `site.github.url`.
 * `author` - Global author information (see below)
 
+If you want to enrich your feed for Feedly, create the following object with any of these optional values:
+
+```yml
+webfeeds:
+  icon: # path to a 96x96 favicon
+  logo: # path to a 30px height svg
+  cover: # path to a large 16:9 banner image
+  accentColor: # hex colour without the hash
+  related: # true = show other stories from your feed under yours
+  analytics: # Google Analytics UA-x tracking ID
+```
+
 ### Already have a feed path?
 
 Do you already have an existing feed someplace other than `/feed.xml`, but are on a host like GitHub Pages that doesn't support machine-friendly redirects? If you simply swap out `jekyll-feed` for your existing template, your existing subscribers won't continue to get updates. Instead, you can specify a non-default path via your site's config.
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -2,7 +2,7 @@
 {% if page.xsl %}
   <?xml-stylesheet type="text/xml" href="{{ '/feed.xslt.xml' | absolute_url }}"?>
 {% endif %}
-<feed xmlns="http://www.w3.org/2005/Atom" {% if site.lang %}xml:lang="{{ site.lang }}"{% endif %}>
+<feed xmlns="http://www.w3.org/2005/Atom"{% if site.webfeeds %} xmlns:webfeeds="http://webfeeds.org/rss/1.0"{% endif %} {% if site.lang %}xml:lang="{{ site.lang }}"{% endif %}>
   <generator uri="https://jekyllrb.com/" version="{{ jekyll.version }}">Jekyll</generator>
   <link href="{{ page.url | absolute_url }}" rel="self" type="application/atom+xml" />
   <link href="{{ '/' | absolute_url }}" rel="alternate" type="text/html" {% if site.lang %}hreflang="{{ site.lang }}" {% endif %}/>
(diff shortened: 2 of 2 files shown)


parkr:
@dylmye Hey! This is neat – I also use Feedly. One concern I have is the longevity of this change. Are these attributes common? It would be helpful to have a universal standard for this that we could rely on so we know it's not a feature we're adding only for Feedly.

dylmye:
Hey @parkr , so this is the (not very well documented) standard Webfeeds. You can see the documentation here: http://webfeeds.org/rss/1.0 . You can see I didn't integrate some fields; these things aren't very well documented. The two links are pretty much what I based this on.

I understand the desire to add features that don't just benefit Feedly but it's pretty difficult to find documentation for enriching data for other RSS readers AFAIK. What I can provide is other projects that have implemented Webfeeds, if that helps:

* [laravel-feed](https://github.com/Laravelium/laravel-feed/blob/master/src/views/rss.blade.php)
* [mastodon](https://github.com/tootsuite/mastodon/blob/main/app/lib/rss_builder.rb)
* [feedrapp](https://github.com/sdepold/feedrapp)

Flipboard doesn't use webfeeds but I don't think we support their media format: https://about.flipboard.com/rss-spec/

dylmye:
bumping this :)

dylmye:
bumping again

README.md
```diff
@@ -34,6 +34,18 @@ The plugin will automatically use any of the following configuration variables,
 * `url` - The URL to your site, e.g., `https://example.com`. If none is provided, the plugin will try to use `site.github.url`.
 * `author` - Global author information (see below)
 
+If you want to enrich your feed for Feedly, create the following object with any of these optional values:
```
parkr, line 37:
```suggestion
If you want to enrich your feed for [Feedly](https://feedly.com), create the following object with any of these optional values:
```

README.md
```diff
@@ -34,6 +34,18 @@ The plugin will automatically use any of the following configuration variables,
...
+
+```yml
+webfeeds:
+  icon: # path to a 96x96 favicon
+  logo: # path to a 30px height svg
+  cover: # path to a large 16:9 banner image
+  accentColor: # hex colour without the hash
+  related: # true = show other stories from your feed under yours
+  analytics: # Google Analytics UA-x tracking ID
```
parkr, line 46: We're generally trying to do away with first-party support for tracking in our products, so maybe we can wait on this one.

README.md
```diff
@@ -34,6 +34,18 @@ The plugin will automatically use any of the following configuration variables,
...
 * `author` - Global author information (see below)
 
+If you want to enrich your feed for Feedly, create the following object with any of these optional values:
+
+```yml
+webfeeds:
+  icon: # path to a 96x96 favicon
+  logo: # path to a 30px height svg
+  cover: # path to a large 16:9 banner image
+  accentColor: # hex colour without the hash
```
parkr, line 44:
```suggestion
  accent_color: # hex colour without the hash ...for what?
```

lib/jekyll-feed/feed.xml
```diff
@@ -39,6 +39,15 @@
     </author>
   {% endif %}
 
+  {% if site.webfeeds %}
+    {% if site.webfeeds.icon %}<webfeeds:icon>{{ site.webfeeds.icon }}</webfeeds:icon>{% endif %}
+    {% if site.webfeeds.logo %}<webfeeds:logo>{{ site.webfeeds.logo }}</webfeeds:logo>{% endif %}
+    {% if site.webfeeds.cover %}<webfeeds:cover>{{ site.webfeeds.cover }}</webfeeds:cover>{% endif %}
+    {% if site.webfeeds.accentColor %}<webfeeds:accentColor>{{ site.webfeeds.accentColor }}</webfeeds:accentColor>{% endif %}
```
parkr, line 46:
```suggestion
    {% if site.webfeeds.accent_color %}<webfeeds:accentColor>{{ site.webfeeds.accent_color }}</webfeeds:accentColor>{% endif %}
```

parkr (commented):
I did a bit of searching around and couldn't find more recent mentions of this than the 2015 Feedly blog post about this, and webfeeds.org is still "Coming soon...". Overall I'm worried about supporting this without Feedly investing more in publishing a standard and speaking about it more. Did I miss something?

dylmye:
Hey sorry that I never caught this.

I totally understand your concerns with adding a poorly documented standard. It would be great to loop in someone from Feedly - I will try to do so.

In the meanwhile, if you're still interested @parkr I'll make the requested changes :)

PR #346, opened 2021-06-19
Allow skipping generation in development

### LINKED ISSUE #331: Disable feed on dev environment
Is there an option to disable jekyll-feed on development environement? Maybe by using ``jekyll.environemnt``?

henrahmagix:
Fix #331

### DIFF
--- a/README.md
+++ b/README.md
@@ -244,6 +244,15 @@ feed:
 
 Note that if you include a tag that is excluded a feed will not be generated for it.
 
+## Skip development
+
+Use `skip_development: true` if you want to turn off feed generation when `jekyll.environment == "development"`, but don't want to remove the plugin (so you don't accidentally commit the removal). Default value is `false`.
+
+```yml
+feed:
+  skip_development: true
+```
+
 ## Contributing
 
 1. Fork it (https://github.com/jekyll/jekyll-feed/fork)
--- a/lib/jekyll-feed/generator.rb
+++ b/lib/jekyll-feed/generator.rb
@@ -8,6 +8,8 @@ module JekyllFeed
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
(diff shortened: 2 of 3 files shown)


README.md
```diff
@@ -244,6 +244,15 @@ feed:
 
 Note that if you include a tag that is excluded a feed will not be generated for it.
 
+## Skip development
+
+Use `skip_development: true` if you want to turn off feed generation when `jekyll.environment == "development"`, but don't want to remove the plugin (so you don't accidentally commit the removal). Default value is `false`.
+
+```yml
+feed:
+  skip_development: true
```
parkr, line 253: I think we could improve this name a bit. What do you think about `disable_in_development` or `skip_in_development`?

lib/jekyll-feed/generator.rb
```diff
@@ -8,6 +8,8 @@ class Generator < Jekyll::Generator
     # Main plugin action, called by Jekyll-core
     def generate(site)
       @site = site
+      return if config["skip_development"] == true && Jekyll.env == "development"
```
parkr, line 11: Let's log in this case so it's clear to the user what's happening. Something like:

```ruby
Jekyll.logger.info "Jekyll Feed:", "Skipping feed generation in development"
```

parkr (commented):
This is a good example of a "power-user" feature – one that most folks won't need but that can help for those with big sites. You just have to remember to set JEKYLL_ENV=production when building your site to push out to the world with this option on!

I think we can iterate on the flag name, otherwise things look good.

parkr:
Fixed by #370!

henrahmagix:
Ah sorry I missed the comments here, thanks so much for your improved replacement @sylhare and thanks for merging @parkr :)

parkr:
No worries, thanks for getting us all started!

Issue #358, opened 2021-12-29
Use Content-Type: application/atom+xml


gsnoff:
Currently the feed generated by this plugin is served with `Content-Type: application/xml`. Some browsers (Firefox on Debian?) tend to treat such content as arbitrary XML, e.g. showing its internal tree, instead of pretty-printing it with a subscribe option, regardless of whether the content itself is valid Atom content with `xmlns` set and whatnot.

When the HTTP header is `Content-Type: application/atom+xml`, the said browsers behave as expected.

parkr:
We use mime-db to generate our mime.types file for Jekyll (and GitHub Pages uses the same). It looks [like `application/atom+xml` requires the `.atom` extension](https://github.com/jshttp/mime-db/blob/562fb6b8a4a7da6c4882490c44441dcaf3e6cc31/db.json#L106-L110).

You can set this with your feed configuration:

```yaml
feed:
  path: /feed.atom
```

parkr:
Unfortunately changing the path globally would cause issues. We could potentially write both paths from now on when `config.feed.path` is unset and set the `rel=alternate` to the new `.atom` path. We can't do 301 redirects (which feed readers generally follow) since GitHub Pages doesn't support redirects. The best bet would be to drop `/feed.xml` in a major version bump and loudly communicate that this path was being moved. Not sure it's feasible for the benefits here when folks can set `config.feed.path` to get this functionality today.

parkr:
Sent a PR to update the docs examples to use `.atom` for customizing the feed path: https://github.com/jekyll/jekyll-feed/pull/359

PR #359, opened 2022-01-02
Use .atom extension in docs examples since we write an Atom feed

parkr:
Using `.atom` also will use the correct MIME type, `application/atom+xml` (https://github.com/jekyll/jekyll-feed/issues/358).

We **cannot** update the default paths to write `.atom` since that would break existing sites (only solution would be to double-write the `.xml` and `.atom` files) so I only updated the `feed.path` configuration where applicable since folks generally copy-paste that kind of thing so this would at least convert some new sites to use `.atom`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,7 @@ Do you already have an existing feed someplace other than `/feed.xml`, but are o
 
 ```yml
 feed:
-  path: atom.xml
+  path: /blog/feed.atom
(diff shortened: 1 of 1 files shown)


mattr- (approved):

ashmaroli:
Thanks @parkr 
@jekyllbot: merge +doc

PR #379, opened 2022-12-24
Allow `categories: true` to make feeds for all categories

### LINKED ISSUE #361: Generate feeds for all categories, instead of only configured categories
To my knowledge it is not currently possible to have feeds automatically be generated for all categories, unlike tags. This appears to have had past requests, which were all closed either automatically or via PRs that only implemented this for tags. I would like to not have to modify `_config.yml` for every new category I add to the Jekyll site, if at all possible.

Is this possible? If not, is there a bounty program where I can put down some money for this to be added?

jez:
### Motivation

Fixes #361

### Summary

This change behaves like how `tags: true` works, and like how @parkr suggested
in the comments of #361.

### Test plan

I've written a test case for this and also verified that it works for my
personal blog when I pass `categories: true` in my `feed:` settings.


jez:
(This is based on the fixed master commit now)

jez:
Sorry about that, I didn’t realize that rubocop was set up for the repo, and I had forgotten to run that locally.

I fixed the rubocop errors by factoring out my changes to a new method. Feel free to suggest any tweaks as you wish.

ashmaroli:
Hello @jez, you needn't apologize for overlooking RuboCop offenses. It's part of the process.
Anyways, I'm finding the use of param name `meta` confusing now that the logic has been abstracted to a separate definition.
How about using `collection` itself:

```ruby
site.collections.each_value do |collection|
  normalize_categories(collection)
end
```

jez:
@ashmaroli I saw you crossed out your request–is there anything else you'd like me to change about this PR?

(otherwise, I think this change is ready to go)

jez:
@ashmaroli friendly bump!

ashmaroli:
Thanks for the ping, @jez. This had fallen off my radar.

ashmaroli (dismissed):
LGTM!

README.md
```diff
@@ -153,6 +153,13 @@ feed:
     - updates
 ```
 
+Or, to generate a feed for all categories:
+
+```yml
+feed:
+  categories: true
```
parkr, line 160: On line 24 of generator.rb, we have special handling for `config["tags"]`. Do we intend to only generate collection-level categories?

README.md
```diff
@@ -153,6 +153,13 @@ feed:
     - updates
 ```
 
+Or, to generate a feed for all categories:
```
parkr, line 156: We might want to explain that in the current implementation, this means that each collection will get its own category feed. This operates differently than tag-based feeds where they unify all documents from any collection into a single feed.

README.md
```diff
@@ -193,6 +200,16 @@ feed:
...
 ```
 
+Or pass `categories: true` to generate a feed for all categories:
+
+```yml
+feed:
+  collections:
+    changes:
+      path: "/changes.atom"
+      categories: true
```
parkr, line 210: Can we have a test for this? It seems like we could over-generate. That is we could generate a category feed which has zero entries since a collection doesn't have any documents with a category that is only present in another collection.

1. Collection "posts" has categories "A" and "B"
2. Collection "archives" has categories "C"
3. Will we generate `/feed/archives/A.xml` with zero entries?

PR #381, opened 2022-12-30
Upgrade to Rake 13 to fix master

jez:
I had a CI run fail for #379 which looks like a failure in the "setup ruby" step.

I'm not sure how the changes I made in #379 could have caused that problem—it fails even before attempting to run the tests, and I haven't touched anything related to dependencies.

I'm trying to figure out whether my intuition is incorrect, or whether CI is actually failing on master. I notice that the last master change was from a couple months ago, so I'm worried that some dependency has broken in the mean time.

### DIFF
--- a/jekyll-feed.gemspec
+++ b/jekyll-feed.gemspec
@@ -22,7 +22,7 @@ Gem::Specification.new do |spec|
 
   spec.add_development_dependency "bundler"
   spec.add_development_dependency "nokogiri", "~> 1.6"
-  spec.add_development_dependency "rake", "~> 12.0"
+  spec.add_development_dependency "rake", "~> 13.0"
   spec.add_development_dependency "rspec", "~> 3.0"
   spec.add_development_dependency "rubocop-jekyll", "~> 0.12.0"
   spec.add_development_dependency "typhoeus", ">= 0.7", "< 2.0"


ashmaroli:
If you look at the log report, you'll see that the error is from gem `sass-embedded` which is a recent entry into our bundle via `jekyll-sass-converter-3.0`.
So, the `master` is most likely affected as well. However, what puzzles me is that the error doesn't occur at repo `jekyll/jekyll` for same Ruby version. Maybe I am overlooking something obvious...

jez:
One thing I noticed when I was working locally (that I managed to deal with separately) was this message:

```
❯ bundle update
Fetching https://github.com/jez/jekyll-feed.git
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies...
sass-embedded-1.57.1-x86_64-linux-gnu requires rubygems version >= 3.3.22, which is incompatible with the current version, 3.1.6
```

I wonder if the bug is that `sass-embedded` made a change that tied them to a specific bundler version, and only later realized that they should declare that dependency in a version bound?

I wonder if it's possible to fix CI by pinning the `sass-embedded` version in CI somehow? I still haven't managed to exactly reproduce the failure as seen by CI in my local environment.

(Thanks for taking a moment to look at this.)

ashmaroli:
That's a different issue @jez. Its asking for a newer version of **Rubygems**, not Bundler.
You can update the version on your local system by running:
```
gem update --system
```

Could you add a commit bumping gem "rake" to use `"~> 13.0"` on this branch? Please ping me after you push so that I can activate the CI run.

parkr:
Interesting. Is Rake 12 incompatible with certain features of Ruby 3? I'm surprised we didn't catch this before.

jez:
@ashmaroli lmk if the latest commit is what you're looking for

ashmaroli:
@parkr I'm not sure if Rake 12.x is incompatible with Ruby 3, but gem `sass-embedded` seems to be using a syntax that is not compatible with Rake 12.
We didn't catch this before because Jekyll Core had moved onto Rake 13 long ago and that `sass-embedded` only became a default member of our bundle / gemset recently.

ashmaroli:
@jez CI is passing.
So your mission has been successful. 🙂

parkr (approved):

parkr:
@jekyllbot: merge +dev

ashmaroli:
@parkr Do we really need Rake for this repository?

PR #386, opened 2023-02-03
enclosure tag for Mailchimp images

Maik93:


lib/jekyll-feed/feed.xml
```diff
@@ -92,6 +92,7 @@
         {% endunless %}
         <media:thumbnail xmlns:media="http://search.yahoo.com/mrss/" url="{{ post_image | xml_escape }}" />
         <media:content medium="image" url="{{ post_image | xml_escape }}" xmlns:media="http://search.yahoo.com/mrss/" />
+        <enclosure url="{{ post_image | xml_escape }}" type="image/jpeg"/>
```
parkr, line 95: Unfortunately the type is not always going to be a JPEG, right?

PR #397, opened 2023-11-14
Add "html_excerpts" flag to allow Atom <summary> tags to contain HTML

Kronopath:
This option should be useful in feeds where the summaries have more complex content or formatting, such as when the post's `description` and `excerpt` front matter options are not set, causing Jekyll to [use the entire first paragraph of the post as the excerpt][0].

It pairs really well with `excerpts_only` for sites which may not want to put the entirety of each post's content in the feed, but who want to preserve the excerpt's rich formatting and content. 

In [my site][1], for example, post excerpts often have images. Its feed currently contains the entire post content, but I'm considering switching it to `excerpts_only`, and I'd very much prefer keeping those images intact in the feed if I do switch.

Note that this passes the W3C validator with no warnings, since feed.xml already uses `type="html"` in its `<summary>` tags. And since it's an option, and not a change in the default behaviour, it shouldn't break existing content.

This PR also comes complete with tests and documentation!

[0]: https://jekyllrb.com/docs/posts/#post-excerpts
[1]: https://www.kronopath.com/blog/

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,16 +207,16 @@ feed:
 The same flag can be used directly in post file. It will disable `<content>` tag for selected post.
 Settings in post file have higher priority than in config file.
 
-## HTML Summaries flag
+## HTML Excerpts flag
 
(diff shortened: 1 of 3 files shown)


Kronopath:
Also I'm aware that this somewhat conflicts with my other changes in #396. Happy to resolve those conflicts later on.

PR #400, opened 2023-12-30
Icon and logo support

Jmuccigr:
A patch to support a blog icon and logo via the _config.yml file. Many feed readers will default to using a site's favicon or icon resources, but the Atom spec supports both an icon (square, so with 1:1 proportions) and a logo (2:1 proportions), so this adds them in. I did this with a more generalizable addition to the xml file:

```yml
blog:
  icon: path/image.png
  logo: path/image2.png
```

In theory other blog properties could be added here. Another way to do it would be the way jekyll already allows a blog title via a `blog_title` entry in the config file, so with `blog_icon` and `blog_logo`, but I think this way is cleaner.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,15 @@ The plugin will automatically use any of the following configuration variables, 
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
 * `url` - The URL to your site, e.g., `https://example.com`. If none is provided, the plugin will try to use `site.github.url`.
 * `author` - Global author information (see below)
+* `blog.icon` - Icon with 1:1 proportions for readers to use for the blog (not supported by all readers; often overridden by a favicon or site icon)
+* `blog.logo` - Logo with 2:1 proportions for readers to use for the blog (not supported by all readers). 
+  For example, both of the preceding can be expressed:
+  ```yml
+  blog:
+    icon: /images/icon.png
+    logo: /images/logo.png
+  ```
+
 
 ### Already have a feed path?
 
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -27,12 +27,20 @@
     <subtitle>{{ site.description | xml_escape }}</subtitle>
   {% endif %}
 
-  {% if site.blog_icon %}
-    {% assign blog_icon = site.blog_icon %}
+  {% if site.blog.icon %}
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -33,6 +33,15 @@ The plugin will automatically use any of the following configuration variables,
 * `description` - A longer description of what your site is about, e.g., "Where I blog about Jekyll and other awesome things"
 * `url` - The URL to your site, e.g., `https://example.com`. If none is provided, the plugin will try to use `site.github.url`.
 * `author` - Global author information (see below)
+* `blog.icon` - Icon with 1:1 proportions for readers to use for the blog (not supported by all readers; often overridden by a favicon or site icon)
+* `blog.logo` - Logo with 2:1 proportions for readers to use for the blog (not supported by all readers). 
+  For example, both of the preceding can be expressed:
+  ```yml
+  blog:
```
parkr, line 40: Hm, I think these should be under the `feed` key, not a `blog` key.

parkr (commented):
Definitely game for this. It needs a test though. The reader I use, NetNewsWire, seems to offer pulling in the favicon but having a custom icon could be neat. 

This was helpful for my understanding of these two fields: https://stackoverflow.com/a/29953980

README.md
Jmuccigr: No objection really, maybe a blog key with “feed-icon” and “feed-logo” sub-keys? I like the idea of keeping the blog info in one place, rather than having multiple keys.

```yml
blog:
  title: My awesome Blog
  feed-icon: /images/icon.png
  feed-logo: /images/logo.png
```

parkr: This plugin uses some global keys like `title` but in general it keeps its own configuration in the `feed` key. So `site.feed.icon` in Liquid. Is there a commonly-used icon key we could use across multiple plugins? If this is only for a feed and would never be reused, then it should be under `feed`. No other Jekyll plugins use the `blog` key, IIRC.

Jmuccigr: I’m not at all familiar with what other plugins do, so I would defer to you. For mY part, my blog has its own title, separate from my site, so I use another key for it, not title. I would also be wary of expecting to share the icon and logo due to their proportion requirements.If feed is already established as a site key, then stick with it, I guess, though personally I distinguish between the blog and the feed.

parkr (commented):
I think this just needs a simple test then we're good to go.

PR #404, opened 2024-01-31
Filter out hidden posts

bkoprucu:
Jekyll doesn't list the posts with a front matter variable `hidden: true` or `hidden: 1` in the paginator / posts listing page.

This change applies the same behavior to the generated feed.

### DIFF
--- a/lib/jekyll-feed/feed.xml
+++ b/lib/jekyll-feed/feed.xml
@@ -50,6 +50,7 @@
   {% unless site.show_drafts %}
     {% assign posts = posts | where_exp: "post", "post.draft != true" %}
   {% endunless %}
+  {% assign posts = posts | where_exp: "post", "post.hidden != true" %}
   {% assign posts = posts | sort: "date" | reverse %}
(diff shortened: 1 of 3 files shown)


guilhermesimoes:
Would be great to have this merged

parkr (changes requested):
Hey, this would match the jekyll-paginate gem. There's also `published: false`. Jekyll's docs don't really describe when to use one vs the other. @mattr- Should we make this align with the pagination plugin?

For now, please update the README here to indicate that posts marked hidden are removed from the feed.

mattr-:
We should match the paginate gem so that things are consistent. I admin that I don't remember why we have `published` and `hidden` properties that seem to do the same thing. Something to investigate later.

iBug:
My understanding is that `published` allows a post to stay in the work tree while not being built at all (i.e. as if it doesn't exist, unless with `--unpublished`). `hidden` allows a post to be built while omitted from listings (i.e. pagination). This PR updates feeds to match the behavior of pagination, as both are generally considered listings.

PR #402, opened 2024-01-07
feat(CI): Add Ruby 3.2/3.3 to the test matrix

Rotzbua:
Add Ruby 3.2 and 3.3 to the test matrix.

PR #411, opened 2024-05-28
4413a.Revert "feat(CI): Add Ruby 3.2/3.3 to the test matrix"

4413a:
Reverts jekyll/jekyll-feed#402

### DIFF
--- a/.github/workflows/ruby.yml
+++ b/.github/workflows/ruby.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


.github/workflows/rubyonrails.yml
parkr, line 1: This is not a rails project.

.github/workflows/ruby.yml
parkr, line 1: This reverts upgrades. We want the upgrades.

parkr (changes requested):

PR #412, opened 2024-05-28
Icon and logo support with simple tests

ChaosKid42:
I took @Jmuccigr's PR #400 and added some simple tests as requested by @parkr in https://github.com/jekyll/jekyll-feed/pull/400#pullrequestreview-1822657054.

PR #414, opened 2024-09-25
Fix style offenses

ashmaroli:
Fixing style issues missed in #412


ashmaroli:
@jekyllbot: merge +dev

PR #421, opened 2025-11-16
Make `image` key configurable for thumbnails/images

Schallbert:
### Why this change?
Some people like me use themes where the `image` key in a post's front matter is either already taken (e.g. for a banner image) or unused. 

I use ([minimal mistakes](https://github.com/mmistakes/minimal-mistakes)) that has a different way of organizing a post's frontmatter with `header.teaser`. 

This PR should solve the issue with a customizable key: `thumb_path_key` can be set in `_config.yml` and will point the liquid template to the correct thumbnail image path. 

This can help deduplicate front matter information.

### What I touched
~~- I have reorganized the generator's code just a bit so its main routine looks shorter for readability.~~
- In the liquid template, `thumb_path_key` is evaluated and split to allow for higher nesting depths in the front matter like `header.teaser`
- I updated the affected readme section with examples, might be a bit lengthy now.

Still this change keeps backwards compatibility due to default values for `image` and `image.path`.

### Testing
- [x]  add tests for changes in `feed.xml`
- [x]  add tests for the updated `.rb` paths


parkr:
Thanks for the PR! It's generally good hygiene to keep each PR narrowly focused. Would you mind pulling your refactor into a separate PR? Anything directly related to the image path change is fine to stay.

Schallbert:
OK, will do. 
I'll also write some tests for the new behavior.

parkr:
Thank you!

Schallbert:
I hope we should be good to go now.

README.md
```diff
@@ -67,7 +67,30 @@ The plugin will use the following post metadata, automatically generated by Jeky
 
 Additionally, the plugin will use the following values, if present in a post's YAML front matter:
 
-* `image` - URL of an image that is representative of the post (can also be passed as `image.path`)
+* `image` - URL of an image that is representative of the post. It can also be passed as `image.path` or set by the user. 
+To customize, go to your `_config.yml` and add the field `thumb_path_key: your.image.key`. Dots express nesting depth of the key.
```
parkr: Did you mean this to be all part of one list item in the unordered list? To create a new paragraph you'll need a blank line between. If intended, no problem.

lib/jekyll-feed/feed.xml
```diff
@@ -115,7 +115,16 @@
         <summary type="html"><![CDATA[{{ post_summary | strip_html | normalize_whitespace }}]]></summary>
       {% endif %}
 
-      {% assign post_image = post.image.path | default: post.image %}
+      {% assign thumb_key = site.data.thumb_key %}
```
parkr: We use the term "image" previously, not "thumb". Could you update this to "image_path_key" or "image_path_frontmatter_key"?

lib/jekyll-feed/feed.xml
```diff
@@ -115,7 +115,16 @@
...
+      {% capture key_parts %}{{ thumb_key }}{% endcapture %}
```
parkr: The key is already a string. Is the capture necessary? Couldn't it be split directly? I believe this is allowing the key to have liquid of its own which might be overkill.

parkr (commented):
Thank you! First pass of review.

README.md
Schallbert: Yes, I wanted it to be as short as possible: 
image

lib/jekyll-feed/feed.xml
Schallbert: OK, changing all ocurrences of "thumb" to "image" for consistency.

lib/jekyll-feed/feed.xml
Schallbert: Right you are. It is a leftover of my tries to get the job done without having to split. Thought I could avoid the loop. Removed.

lib/jekyll-feed/generator.rb
```diff
@@ -12,6 +12,7 @@ def generate(site)
         Jekyll.logger.info "Jekyll Feed:", "Skipping feed generation in development"
         return
       end
+      site.data['image_key'] = get_image_key
```
parkr: I believe this can be removed, is that right? I don't think we're using it in the feed.xml Liquid.

lib/jekyll-feed/feed.xml
```diff
@@ -115,7 +115,15 @@
         <summary type="html"><![CDATA[{{ post_summary | strip_html | normalize_whitespace }}]]></summary>
       {% endif %}
 
-      {% assign post_image = post.image.path | default: post.image %}
+      {% assign image_key = site.data.image_key %}
```
parkr: In a site with `site.data.image_key` missing, will any of this fail? Perhaps we should consider adding a `default: "image"` here so that it always has an `image_key`. Then we can remove the `default: post.image` below on line 126.

lib/jekyll-feed/generator.rb
```diff
@@ -53,6 +54,14 @@ def feed_path(collection: "posts", category: nil)
       collections.dig(collection, "path") || "#{prefix}.xml"
     end
 
+    def get_image_key
+      image_path_key = @site.config.fetch('image_path_key', "image.path")
+      if image_path_key != "image.path"
+        Jekyll.logger.info "Jekyll Feed:", "Using custom image_path_key = #{image_path_key}"
+      end
+      return image_path_key
+    end
```
parkr, line 62: Same with this -- I believe it can be removed.

lib/jekyll-feed/generator.rb
Schallbert: Currently this is where I hand over the image key to `feed.xml` so it becomes available for retrieving the image path. 
If we pulled the key from the config directly in Liquid, we could avoid adding code in `ruby` altogether. Without `ruby` involved though, I wouldn't  know how to add the log line about using a custom path key.

lib/jekyll-feed/feed.xml
Schallbert: It wouldn't fail because `site.data.image_key` defaults to `image.path` which tries replicating the behavior we had before the change (if `image.path` is empty it tries `post.image`)

lib/jekyll-feed/generator.rb
Schallbert: As written above. Would you like me to try a liquid-only variant? It would be  getting the key from the config `posts_limit` times then and we wouldn't have a log line, but it will definetely simplify things.

Schallbert:
- `generator.rb`: Modifiy `get_image_key` to `check_custom_image_key` that won't to anything but log if we detect a custom image key
- `generator_spec.rb`: Removed altogether as there's nothing left in the generator that affects the feed.
- `feed.xml` Read `site.image_path_key` directly.

lib/jekyll-feed/feed.xml
```diff
@@ -115,7 +115,15 @@
...
       {% endif %}
 
-      {% assign post_image = post.image.path | default: post.image %}
+      {% assign image_key = site.image_path_key | default: "image.path" %}
+      {% assign key_array = image_key | split: "." %}
+      
+      {% assign image_value = post %}
+      {% for key in key_array %}
+        {% assign image_value = image_value[key] %}
+      {% endfor %}
```
Schallbert: I just see that these lines could be moved above the "post" loop as they just need to be executed once which would make the code more efficient. On the other hand readability and cohesiveness would degrade, as the variable `image_value` is instantiated there but used somewhere else. Any opinions?

lib/jekyll-feed/generator.rb
parkr: Let's keep the log line - I think it's useful.

spec/jekyll-feed_spec.rb
```diff
@@ -98,6 +98,31 @@
...
+  context "custom image_key" do
+    image_key_tests = [
+      { thumbs: 'header.image.path', expected: 'http://example.org/image.png' }, 
+      { thumbs: 'thumb.path', expected: 'https://cdn.example.org/absolute.png?h=188&amp;w=250' }, 
+      { thumbs: 'thumbnail', expected: 'http://example.org/object-image.png' },
+    ]
+
+    image_key_tests.each do |test_case|
+      it "correctly processes image_key: '#{test_case[:thumbs]}'" do
+        site.data["image_key"] = test_case[:thumbs]
```
parkr: Does this need to be `image_path_key` to pass?

lib/jekyll-feed/feed.xml
parkr: The `image_key` and `image_key_array` could be pre-calculated above outside the loop to speed things up. The `image_value` would need to be calculated inside the loop to ensure it gets the right value for each post. I think the cleanest solution could be:

Outside the loop: if site.image_path_key is non-null, build key_array.
Inside the loop: if key_array is non-null, build image_value. Assign post_image as `post_image = image_value | default: post.image.path | default: post.image`

What do you think?

Schallbert:
> Outside the loop: if site.image_path_key is non-null, build key_array.

Makes sense to me. What I can't get my head around right now (too tired) is how I should build the value outside the loop and prepend `post` afterwards inside. Maybe I need to work with empty array index [0] , place `post` and reconvert it to a key or maybe I need the `post` array of length 1 and append the pre-build `image_value`...

Will take another look at it tomorrow.

Schallbert:
Now, three tests fail:
image
They all fail because at a place where I'd expect `object-image.png` the spec result shows `image.png` and I just cannot figure out why. My real website is built fine and has no image-swap issues.
image
I guess I need some help getting the tests to pass.

